PR #117, opened 2014-04-05
Proguard will obfuscate class names.

kkocel:
Previous proguard example left class names unchouched.


JakeWharton:
I merged #161.

PR #140, opened 2014-05-18
Adding InjectResource to the annotation processor

vignesh-iopex:
Added Resource Binding module that supports the following data types String, Drawable & Animation.


JakeWharton:
See #244

PR #203, opened 2014-12-03
Performance tuning for finding injector/resettor

nguyentruongtho:
Instead of recursively visit every class in the class inheritance hierarchy twice for both calls `inject` & `reset`, do it only one.

Note that there are few ways to do [this](https://github.com/nguyentruongtho/butterknife/blob/3f03ae41c6066ace1eeed3e19b900db3b61d9731/butterknife/src/main/java/butterknife/ButterKnife.java#L293-293) but the selected one is considered to be most suitable (generated class always has a naked constructor) and fastest, according to my dummy benchmark [here](https://gist.github.com/nguyentruongtho/b221292d4a52baed0ff2)


JakeWharton:
Squashed and merged (with extremely tiny tweaks). Thanks!

Issue #212, opened 2015-01-08
Allow interface types for injected views


emilsjolander:
Hi, i ran into a slight issue using butterknife just now and thought i would open an issue to discuss it before making a pull request. I would like the ability to inject views into fields declared by just an interface type. Currently butterknife makes sure the type declared is a descendant of `android.view.View`.

A quick example illustrating the issue i ran into which i think a change like this would fix:
```java
public class CheckableLinearLayout extends LinearLayout implements Checkable {

    @InjectView(R.id.checkable)
    Checkable checkable;

    public CheckableLinearLayout(Context context, AttributeSet attrs) {
        super(context, attrs);
    }

    @Override
    protected void onFinishInflate() {
        super.onFinishInflate();
        ButterKnife.inject(this);
    }

    @Override
    public void setChecked(boolean checked) {
        checkable.setChecked(checked);
    }

    @Override
    public boolean isChecked() {
        return checkable.isChecked();
    }

    @Override
    public void toggle() {
        checkable.toggle();
    }

}
```
This code currently does not compile because butterknife validates that the field annotated with `@InjectView` is a ´android.view.View´ subclass which it is not.

Checkable is just one example of a time when you want to inject a view with a certain type of behaviour but don't want to rely on the actual implementation. The implementation would only be specified in the xml file making it easy to change implementation depending on configurations.

I'm happy to implement this and submit a pull request if you think the idea is good. In that case my plan is to allow either subclasses of ´android.view.View´ or interfaces as declarations of the injected field. Butterknife would still complain if a concrete implementation not subclassing ´android.view.View´ was declared.

JakeWharton:
Yep. Let's kill the check.

JakeWharton:
Let's also catch `ClassCastException` and re-throw it with a message that includes the ID and field name, similar to the message for when a view is `null`.

emilsjolander:
Sounds good, i'll submit a pull as within the next few days

JakeWharton:
Great. Thanks!

PR #214, opened 2015-01-12
Allow interface types for injected views

emilsjolander:
This pull request implements what was talked about in issue #212. In summary it allows any injected field to not only be a view subclass but can also be declared with an interface type.

This pull request also adds  a custom error message for when casting injected views fails. I'm currently slightly uncertain if the direction i went with this is good so i'm very open to suggestions. Currently listener parameters are not cast in this way but if this direction looks good i'll update the pull request so that the listener parameters are cast in a similar way.


butterknife/src/main/java/butterknife/internal/ViewInjector.java
```diff
@@ -129,7 +129,9 @@ private void emitCollectionBinding(StringBuilder builder, CollectionBinding bind
         builder.append(',');
       }
       builder.append("\n        ");
-      emitCastIfNeeded(builder, binding.getType());
+      if (!binding.getType().equals(VIEW_TYPE)) {
```
JakeWharton: Can we extract this onto `binding` as a `requiresCast()` method? I don't like the duplicated logic all over.

JakeWharton:
This looks great. I just have one small suggestion. If you can't get to it in a day or so I'll just make the change myself and merge.

emilsjolander:
@JakeWharton fixed in https://github.com/emilsjolander/butterknife/commit/334000d4df35dec345d94180b66f15364b730a80

emilsjolander:
Added similar casting code for listener parameters. Should be ready to merge now unless you have any comments on the most recent changes.

emilsjolander:
My local setup doesn't seem to test the same tests as travis is set up to do. I'll fix the issues it found.

JakeWharton:
I'm going to steal this change from you, rebase it on master, and see if I can clean up some of the generated code (I don't like the double method dispatch). I'll squash and amend to your changes so you retain the commit author.

emilsjolander:
cool, sounds good!

PR #220, opened 2015-01-28
Clarified reset section a bit

jtietema:
I think the description incorrectly stated fragment, when it should have been views.


JakeWharton:
This whole sentence is wonky. I'm going to change it up completely.

PR #222, opened 2015-02-01
Added AndroidLibs shield

Cyruxx:
Added a ShieldIO Shield for the AndroidLibs listing of ButterKnife. :)


JakeWharton:
Sorry but I see no value in these things. You're already on the GitHub page for the project, why would you want to go somewhere else that just tells you the same stuff (but less)?

Cyruxx:
Well, to find other cool libraries. :-)

JakeWharton:
There's a few of these sites. Adding one would mean that I would have to add them all, or figure out how to justify which ones were worth having. Ultimately it has nothing to do with the library directly so blanket reject anything like this is easier on me.

PR #234, opened 2015-03-03
fix documentation for @OnTouch

runningcode:


butterknife/src/main/java/butterknife/OnTouch.java
```diff
@@ -13,7 +13,7 @@
 /**
  * Bind a method to an {@link OnTouchListener OnTouchListener} on the view for each ID specified.
  * <pre><code>
- * {@literal @}OnTouch(R.id.example) void onTouch() {
+ * {@literal @}OnTouch(R.id.example) boolean onTouch() {
  *   Toast.makeText(this, "Touched!", LENGTH_SHORT).show();
```
JakeWharton: Needs a return value as well then!

runningcode: there we go!

JakeWharton:
Thanks!

PR #242, opened 2015-03-14
Added @OnAutoCompleteItemClick annotation to bind onItemClickListener on...

manijshrestha:
... AutoCompleteTextView


manijshrestha:
@JakeWharton , I am wondering if "targetType" could be changed to be an array? This annotation and OnRadioCheckedChanged are basically same as @OnItemSelected and @CheckedChanged?  Any thoughts?

maulik-dodia:
I'm not getting this new Annotation while implementing butter knife for AutoCompleteTextView. It is not there in suggestion.

manijshrestha:
This never got merged. It was written a while back and now has merge conflict. You might want to do it in code instead for now.

PR #244, opened 2015-03-15
Support binding resource types to fields.

JakeWharton:


gabrielittner:
On API level 22 `getDrawable` should probably be called directly on `Context`.

https://plus.google.com/u/0/+AlanViverette/posts/YHy5drbz3oN

JakeWharton:
I'm aware as the javadoc states such. It was 4am and I was tired and opted not to bother yet.

PR #267, opened 2015-05-25
[DISCUSSION] Add Setter Binding to FindView

exallium:
Allows annotation of a setter method with @FindView.

I'm wondering if there's any value in this.  I'd like to be able to use Butterknife from Kotlin using KAPT.  That syntax would look like:

```kotlin
var myButton: Button? = null
    [FindView(R.id.myButton)] set
```

I realize that ```KotterKnife``` is a thing, and perhaps it's a better use-case for this (heck, I'd like to see an entire Dependency Injection framework done this way) but I thought I'd run this by you, given I was playing around with it anyhow.

This is not created to be merged, this is simply a discussion point.  I am willing to put the effort into writing the rest of the implementation if this is deemed a worthy venue, I just didn't want to do all the work before getting feedback.

### DIFF
--- a/butterknife-sample/src/main/java/com/example/butterknife/SimpleActivity.java
+++ b/butterknife-sample/src/main/java/com/example/butterknife/SimpleActivity.java
@@ -29,11 +29,17 @@ public class SimpleActivity extends Activity {
     }
   };
 
-  @FindView(R.id.title) TextView title;
   @FindView(R.id.subtitle) TextView subtitle;
   @FindView(R.id.hello) Button hello;
   @FindView(R.id.list_of_things) ListView listOfThings;
   @FindView(R.id.footer) TextView footer;
+
+  TextView title;
+
+  @FindView(R.id.title)
+  void setTitle(TextView title) {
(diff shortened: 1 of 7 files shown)


JakeWharton:
I'm skeptical for two reasons:
 * The features has little-to-no use in Java land. In fact, I can't think of a case where I would use this over just doing field binding.
 * Butter Knife is a boilerplate reduction library and the only reason it uses code generation is so that there's no real overhead (and to poke fun at RoboGuice people switching to Dagger and complaining about the lack of "view injection", whatever the hell that is). With Kotlin, however, the language is much more expressive and it allows you to do things like Kotter Knife as a way to reduce the boilerplate and not have any (real) overhead.

exallium:
I agree.  (FWIW I'm using KotterKnife in a side project and have read / modified it as well).

Delegated Properties are a much more natural solution to the problem solved by ButterKnife.

With that, we can close this.  Keeping the library focused on Java and keeping it simple sounds better than adding in some functionality that has no use in Java itself.

Thanks.

PR #270, opened 2015-06-07
check if enclosingElement is in default package

xesam:


xesam:
check if enclosingElement is in default package

JakeWharton:
Do we actually need this? You cannot use classes from the default package in anything but the default package. Are you writing all your Android app's classes in the default package?

JakeWharton:
Tentatively closing, but will reconsider if you make a compelling argument.

PR #278, opened 2015-06-28
Fix website code for view lists.

lukaciko:
Diamond operator cannot be used with anonymous inner classes, the third parameter of `ButterKnife.apply` with `View.Alpha` has Float type and therefore accepts float and not int.


JakeWharton:
Yep yep. Thanks!

PR #288, opened 2015-06-30
Remove explicit array allocation by leaning on copy.

### LINKED ISSUE #287: Finder.arrayOf(...) always returns Object[] if an optional view is null
If you bind an array of views and one is null/optional, `Finder.arrayOf(…)` returns an instance of `Object[]`, which results in a `ClassCastException` at runtime.

This worked fine before 7.0.0 when we used `@InjectViews`, I'm not sure what changed under the hood. It might be the case you shouldn't try to have nullable view in an array anyway, but just wanted to let you know in case this wasn't intentional.

### Example

```java
@Nullable
@Bind({R.id.textview0, R.id.textview1, R.id.textview2})
TextView[] textViews;
```

Results in this crash at runtime:

```
Caused by: java.lang.ClassCastException: java.lang.Object[] cannot be cast to android.widget.TextView[]
            at io.sweers.test.MainActivity$$ViewBinder.bind(MainActivity$$ViewBinder.java:11)
            at io.sweers.test.MainActivity$$ViewBinder.bind(MainActivity$$ViewBinder.java:8)
            at butterknife.ButterKnife.bind(ButterKnife.java:328)
            at butterknife.ButterKnife.bind(ButterKnife.java:246)
            at io.sweers.test.MainActivity.onCreate(MainActivity.java:23)
```

Full example here: https://github.com/hzsweers/Butterknife-test

JakeWharton:
Closes #287.

### DIFF
--- a/butterknife/src/main/java/butterknife/ButterKnife.java
+++ b/butterknife/src/main/java/butterknife/ButterKnife.java
@@ -9,6 +9,7 @@ import android.util.Log;
 import android.util.Property;
 import android.view.View;
 import butterknife.internal.ButterKnifeProcessor;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -114,24 +115,14 @@ public final class ButterKnife {
     };
 
     private static <T> T[] filterNull(T[] views) {
-      int newSize = views.length;
-      for (T view : views) {
-        if (view == null) {
-          newSize -= 1;
-        }
-      }
-      if (newSize == views.length) {
-        return views;
-      }
-      //noinspection unchecked
-      T[] newViews = (T[]) new Object[newSize];
-      int nextIndex = 0;
-      for (T view : views) {
+      int end = 0;
+      for (int i = 0; i < views.length; i++) {
+        T view = views[i];
         if (view != null) {
-          newViews[nextIndex++] = view;
+          views[end++] = view;
         }
       }
-      return newViews;
+      return Arrays.copyOfRange(views, 0, end);
     }
 
     public static <T> T[] arrayOf(T... views) {


hzsweers:
LGTM!

butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -114,24 +115,14 @@ private ButterKnife() {
...
-      //noinspection unchecked
-      T[] newViews = (T[]) new Object[newSize];
-      int nextIndex = 0;
-      for (T view : views) {
+      int end = 0;
+      for (int i = 0; i < views.length; i++) {
+        T view = views[i];
         if (view != null) {
-          newViews[nextIndex++] = view;
+          views[end++] = view;
```
JakeWharton: I'd be nice to avoid this case of re-assigning the same reference in the happy path (no nulls).

hzsweers: Well, a one liner change to the existing implementation would be to change 127 to what `Arrays.copyOfRange` does under the hood:

```java
T[] newViews = Array.newInstance(views.getClass().getComponentType(), newSize);
```

When I think about it, I don't know which option would really be better in this case though. I'd speculate that the average case (where most, if not all the views are non-null) is faster in the original implementation using `Array.newInstance`. I'm not sure what point the `Arrays.copyOfRange` implementation would become faster, but honestly think either case would be fine perf-wise. Anyone with enough views to cause a noticeable difference probably has bigger problems.

JakeWharton: Yep. Agreed. Just [released][1]!

 [1]: http://repo1.maven.org/maven2/com/jakewharton/butterknife/7.0.1/

hzsweers: 🎉

PR #290, opened 2015-07-02
Fix Javadoc

ytRino:
trivial fix.
Rename resource binding classes and R._integer*_ class


JakeWharton:
> Rename resource binding classes and R.integer* class

Use this as the commit message next time.

ytRino:
ah, sry. and thank you for advice.

PR #296, opened 2015-07-06
Javadoc cleanup

tomxor:


JakeWharton:
This is what happens when you change the name multiple times!

Issue #300, opened 2015-07-07
Feature Request: Annotation for binding array resources


tpfaff:
Would love to be able to bind array resources.

@BindString binds an R.string ID to a String field.

@BindArray binds an R.array ..........

tomxor:
`@BindStringArray` was introduced in #297 and should probably be available on the next release. 

@JakeWharton there's also [getIntArray(int)](http://developer.android.com/reference/android/content/res/Resources.html#getIntArray(int)), [obtainTypedArray(int)](http://developer.android.com/reference/android/content/res/Resources.html#obtainTypedArray(int)) and [getTextArray(int)](http://developer.android.com/reference/android/content/res/Resources.html#getTextArray(int)) -- would you rather have multiple annotations (i.e. `@BindIntArray`, `@BindTypedArray`, `@BindTextArray`) - or switch to `@BindArray` and figure out which method to call according to the type of enclosing element?

tpfaff:
Awesome!

JakeWharton:
Oops. I thinking switching on `@BindArray` would be ideal provided there's no change of ambiguity.

JakeWharton:
Although it prevents us from using the support-annotations in the future #144.

tomxor:
Can't we use `@ArrayRes`?

JakeWharton:
Oh it's not typed? I guess that makes sense (but is unfortunate). I'm not an array resources user (hence its omission and my lack of knowledge here). If that's true, then go for it!

tomxor:
:+1:

PR #301, opened 2015-07-07
Support binding array resources (String, int, text and TypedArray).

tomxor:
This replaces `@BindStringArray` with `@BindArray` in order to support multiple array types. The type of annotated element will dictate which method of the `Resources` class to invoke in order to fetch the array.


tomxor:
Fixes #300

butterknife/src/main/java/butterknife/internal/ButterKnifeProcessor.java
```diff
@@ -644,22 +645,24 @@ private void parseResourceString(Element element, Map<TypeElement, BindingClass>
...
 
-    // Verify that the target type is String[].
-    if (!"java.lang.String[]".equals(element.asType().toString())) {
-      error(element, "@%s field type must be 'String[]'. (%s.%s)",
-          BindStringArray.class.getSimpleName(), enclosingElement.getQualifiedName(),
+    // Verify that the target type is supported.
+    String methodName = getArrayResourceMethodName(element);
+    if (methodName == null) {
+      error(element,
+          "@%s field type must be one of: " + "'String[], int[], CharSequence[], %s'. (%s.%s)",
```
JakeWharton: needless `+` between string contstants

butterknife/src/main/java/butterknife/internal/ButterKnifeProcessor.java
```diff
@@ -667,15 +670,37 @@ private void parseResourceStringArray(Element element,
...
     erasedTargetNames.add(enclosingElement.toString());
   }
 
+  /**
+   * Returns a method name from the {@link android.content.res.Resources} class for array resource
+   * binding, null if the element type is not supported.
+   */
+  private String getArrayResourceMethodName(Element element) {
+    TypeMirror typeMirror = element.asType();
+    if (TypeKind.ARRAY.equals(typeMirror.getKind())) {
```
tomxor: This can be simplified to simply match the type against `String[]`, `int[]`, etc. Not sure which way is better ¯\_(ツ)_/¯

butterknife/src/main/java/butterknife/internal/ButterKnifeProcessor.java
```diff
@@ -667,15 +670,37 @@ private void parseResourceStringArray(Element element,
...
     bindingClass.addResource(binding);
 
     erasedTargetNames.add(enclosingElement.toString());
   }
 
+  /**
+   * Returns a method name from the {@link android.content.res.Resources} class for array resource
+   * binding, null if the element type is not supported.
+   */
+  private String getArrayResourceMethodName(Element element) {
```
JakeWharton: +`static`

butterknife/src/main/java/butterknife/internal/ButterKnifeProcessor.java
```diff
@@ -667,15 +670,37 @@ private void parseResourceStringArray(Element element,
...
+      ArrayType arrayType = (ArrayType) typeMirror;
+      String componentType = arrayType.getComponentType().toString();
+      if ("java.lang.String".equals(componentType)) {
+        return "getStringArray";
+      } else if ("int".equals(componentType)) {
+        return "getIntArray";
+      } else if ("java.lang.CharSequence".equals(componentType)) {
+        return "getTextArray";
+      }
+    } else if (TYPED_ARRAY_TYPE.equals(typeMirror.toString())) {
```
JakeWharton: I'd prefer if this method was written as:
```java
if (typed array) {
  return "obtain..";
}
if (array) {
 ...
}
return null;
```
to avoid the so-called dangling else which requires me to backtrack and figure out what it's behavior is.

butterknife/src/test/java/butterknife/internal/BindArrayTest.java
```diff
@@ -0,0 +1,166 @@
...
+        "  @BindArray(1) String one;",
+        "}"
+    ));
+
+    ASSERT.about(javaSource()).that(source)
+        .processedWith(new ButterKnifeProcessor())
+        .failsToCompile()
+        .withErrorContaining(
+            "@BindArray field type must be one of: 'String[], int[], CharSequence[], "
+                + "android.content.res.TypedArray'. (test.Test.one)")
```
JakeWharton: I don't think the single quotes in this error message add anything. Drop 'em?

JakeWharton:
Looks great. A couple small nits.

tomxor:
:+1: thanks!

PR #304, opened 2015-07-08
Suppress unexpected exception in Previewer

kevinmost:
Required views that were not found were calling a method that throws an
UnsupportedOperationException when in the Android Studio Previewer
(due to Previewer's use of BridgeResources). A check is now performed
to ensure that the view is not in Edit Mode before calling this method.

### DIFF
--- a/butterknife/src/main/java/butterknife/ButterKnife.java
+++ b/butterknife/src/main/java/butterknife/ButterKnife.java
@@ -94,6 +94,15 @@ public final class ButterKnife {
       @Override public Context getContext(Object source) {
         return ((View) source).getContext();
       }
+
+      @Override protected String getResourceEntryName(Object source, int id) {
+        final View view = (View) source;
+        // In edit mode, getResourceEntryName() is unsupported due to use of BridgeResources
+        if (view.isInEditMode()) {
+          return "<unavailable while editing>";
+        }
+        return super.getResourceEntryName(source, id);
+      }
     },
     ACTIVITY {
       @Override protected View findView(Object source, int id) {
@@ -136,7 +145,7 @@ public final class ButterKnife {
     public <T> T findRequiredView(Object source, int id, String who) {
       T view = findOptionalView(source, id, who);
       if (view == null) {
-        String name = getContext(source).getResources().getResourceEntryName(id);
+        String name = getResourceEntryName(source, id);
         throw new IllegalStateException("Required view '"
             + name
             + "' with ID "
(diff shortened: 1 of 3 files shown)


butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -94,6 +94,15 @@ private ButterKnife() {
       @Override public Context getContext(Object source) {
         return ((View) source).getContext();
       }
+
+      @Override protected String getResourceEntryName(Object source, int id) {
+        final View view = (View) source;
+        // In edit mode, getResourceEntryName() is unsupported due to use of BridgeResources
+        if (view.isInEditMode()) {
+          return "{ Names cannot be retrieved while in Edit Mode }";
```
JakeWharton: This is a long string. How about `"<name unavailable>"`?

JakeWharton: Or `"<unavailable while editing>"`

butterknife/src/test/java/butterknife/ButterKnifeTest.java
```diff
@@ -152,4 +154,16 @@ public void resetViewsCache() {
           + " for yo mama was not found. If this view is optional add '@Nullable' annotation.");
     }
   }
+
+  @Config(shadows = EditModeShadowView.class)
+  @Test public void finderThrowsLessNiceErrorInEditMode() {
+    View view = new View(Robolectric.application);
+    try {
+      ButterKnife.Finder.VIEW.findRequiredView(view, android.R.id.button1, "yo mama");
```
serj-lotutovici: May be adding `fail()` after calling `findRequiredView` could be a possibility. Otherwise the test wont do anything if the exception is not thrown.

JakeWharton: :+1: good eye

JakeWharton:
One thing to fix, and then can you squash to one commit?

kevinmost:
Sounds good, just did that. I also added the `fail()` to the existing test that was there (the one that's basically the same test, but not for Edit Mode). If that doesn't sound right, let me know.

butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -94,6 +94,15 @@ private ButterKnife() {
       @Override public Context getContext(Object source) {
         return ((View) source).getContext();
       }
+
+      @Override protected String getResourceEntryName(Object source, int id) {
+        final View view = (View) source;
+        // In edit mode, getResourceEntryName() is unsupported due to use of BridgeResources
```
JakeWharton: I filed http://b.android.com/181872 for this to be fixed.

PR #316, opened 2015-07-14
Support `@BindBitmap` annotation

tomxor:
#315

Issue #315, opened 2015-07-14
@BindBitmap(R.drawable.foo) Bitmap foo


JakeWharton:
```java
foo = BitmapFactory.decodeResource(res, R.drawable.foo);
```

tomxor:
Is this going to replace `@BindDrawable`?

JakeWharton:
No. Different use cases. Sometimes you need a raw Bitmap (notification
icon, task description icon, etc.).

tomxor:
Should we actually clear the `Bitmap` reference in `unbind()`?

JakeWharton:
I don't think so. Other resources do not behave this way and while there is a native component, the point of unbind it to prevent leaking activities from fragments when you can't really leak a Bitmap in the same fashion.

vsay01:
I use ButterKnife 7.0.1 and I need to use the Bitmap resource, According to this thread @BindBitmap annotation was added, but when it is not found. Is it removed? Any other way to achieve this goal?

JakeWharton:
It has not been released yet.

PR #317, opened 2015-07-14
Clear Bitmap references on unbind().

tomxor:
#315, #316 

Not sure whether or not this is desirable, so feel free to reject.


JakeWharton:
I don't think so. I think it would be more surprising than helpful: https://github.com/JakeWharton/butterknife/issues/315#issuecomment-121392839

PR #210, opened 2015-01-06
Split the compiler and runtime into separate artifacts.

riotopsys:

Issue #192, opened 2014-10-30
Split the compiler and runtime into separate artifacts.


JakeWharton:
Coalescing #103, #182, and #191.

huangcd:
Hi Jake, Any fixed or workaround for this issues? I'm facing the same problem on Android 5.0.1

felipecsl:
#210 should fix this

JakeWharton:
This happened.

PR #323, opened 2015-07-31
Split the compiler and runtime into separate artifacts.

serj-lotutovici:
Posible fix for #192 
Didn't see #210 before it was too late (but the PR was not updated for a longer time, so I figured: why not!? :) )


butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -8,25 +8,26 @@
...
-import static butterknife.internal.ButterKnifeProcessor.JAVA_PREFIX;
+import static butterknife.internal.InternalKeys.ANDROID_PREFIX;
+import static butterknife.internal.InternalKeys.BINDING_CLASS_SUFFIX;
+import static butterknife.internal.InternalKeys.JAVA_PREFIX;
 
 /**
  * Field and method binding for Android views. Use this class to simplify finding views and
  * attaching listeners by binding them with annotations.
- * <p>
+ * <p/>
```
JakeWharton: Revert. This is invalid HTML3.2 which is what Javadoc uses. If you compile with JDK 8 it doclint will actually fail.

butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -8,25 +8,26 @@
...
  * Finding views from your activity is as easy as:
  * <pre><code>
  * public class ExampleActivity extends Activity {
  *   {@literal @}Bind(R.id.title) EditText titleView;
  *   {@literal @}Bind(R.id.subtitle) EditText subtitleView;
- *
+ * <p/>
```
JakeWharton: This isn't between a paragraph.

butterknife/src/main/java/butterknife/internal/InternalKeys.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Contains shared constants between the annotation processor and {@link butterknife.ButterKnife}
+ */
+public final class InternalKeys {
+  public static final String BINDING_CLASS_SUFFIX = "$$ViewBinder";
+  public static final String ANDROID_PREFIX = "android.";
+  public static final String JAVA_PREFIX = "java.";
+
+  /** Prevent class from instantiation (aka being paranoid) */
```
JakeWharton: Delete comment.

butterknife/src/main/java/butterknife/internal/InternalKeys.java
```diff
@@ -0,0 +1,15 @@
...
+  private InternalKeys() {
+    throw new AssertionError("InternalKeys is a constant container. Should not be instantiated");
```
JakeWharton: Typical message here is just "No instances."

butterknife/pom.xml
```diff
@@ -60,7 +54,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <configuration>
-          <excludePackageNames>butterknife.internal:butterknife.internal.*</excludePackageNames>
```
JakeWharton: This is still needed to ensure `InternalKeys` doesn't show up in Javadoc.

butterknife/pom.xml
```diff
@@ -52,7 +47,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <compilerArgument>-proc:none</compilerArgument>
         </configuration>
       </plugin>
```
JakeWharton: You can delete this whole `<plugin>`

butterknife/src/main/java/butterknife/ButterKnife.java
JakeWharton: (all of them)

butterknife-compiler/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+      <artifactId>fest-assert-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.squareup</groupId>
+      <artifactId>fest-android</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.robolectric</groupId>
```
JakeWharton: This dependency can probably be removed

JakeWharton:
This looks fantastic! I have only a few minor comments, then let's get this merged!

butterknife-compiler/pom.xml
serj-lotutovici: Unfortunately not. `AllTheThingsTest` and `OnPageChangeTest` indirectly require it. Removing it will result in test failure do compilation error after the `ButterKnifeProcessor` execution.

serj-lotutovici:
@JakeWharton fixed issues, based on your comments. Bay chance do you have any "code style" profile file for intellij idea, that I can use? It seams that my ide settings profile defers from your, which can cause same formatting issue in the future. Thx in advance :)

tomxor:
@serj-lotutovici https://github.com/square/java-code-styles

JakeWharton:
Squashed and merged.

JakeWharton:
Thanks for this!

PR #344, opened 2015-08-28
linkedhashmap isn't necessary

runningcode:
I couldn't find an instance where order matters or the hash map was iterated over.

Also, depending on the number of entries, I believe ArrayMap could be more memory efficient in certain cases.


JakeWharton:
We almost always use LinkedHashMap unconditionally unless there's a reason not to (like needing a concurrent map). The advantage is that they are _always_ deterministic in behavior which allows reproducibility of bugs and determinism in tests. You aren't going to remember to update a map's type for this functionality when you start iterating over it.

As to ArrayMap, we don't care about memory here, only speed. The O(1) behavior is more desirable.

PR #354, opened 2015-09-17
Press more juice

pyricau:
Remove extra waste of string comparison when going through bind class hierarchy

### DIFF
--- a/butterknife/src/test/java/butterknife/ButterKnifeTest.java
+++ b/butterknife/src/test/java/butterknife/ButterKnifeTest.java
@@ -138,11 +138,11 @@ public class ButterKnifeTest {
     assertThat(ButterKnife.BINDERS).contains(entry(Example.class, ButterKnife.NOP_VIEW_BINDER));
   }
 
-  @Test public void bindingKnownPackagesIsNoOp() {
+  @Test public void bindingKnownPackagesCaches() {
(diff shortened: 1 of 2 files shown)


pyricau:
This only improves the situation if you get several miss that all go up until the framework class.

JakeWharton:
Test needs updated

pyricau:
Updated, tests fixed.

butterknife/src/test/java/butterknife/ButterKnifeTest.java
```diff
@@ -138,11 +138,11 @@ public void resetViewsCache() {
     assertThat(ButterKnife.BINDERS).contains(entry(Example.class, ButterKnife.NOP_VIEW_BINDER));
   }
 
-  @Test public void bindingKnownPackagesIsNoOp() {
+  @Test public void bindingKnownPackagesCaches() {
     ButterKnife.bind(new Activity());
-    assertThat(ButterKnife.BINDERS).isEmpty();
+    assertThat(ButterKnife.BINDERS).hasSize(1);
```
JakeWharton: This needs to assert the no-op binding was the one which was cached.

JakeWharton:
Fixed as e9c1533e2af2b26bf26ce13c139ae6ffdeb3a7c5

PR #357, opened 2015-09-22
Add action and setter methods that take single view

Orbar:
This extends the concept of `Action` and `Setter` and make them available to single views while maintaining the same interface for actual action

### DIFF
--- a/butterknife/src/main/java/butterknife/ButterKnife.java
+++ b/butterknife/src/main/java/butterknife/ButterKnife.java
@@ -9,7 +9,6 @@ import android.util.Log;
 import android.util.Property;
 import android.view.View;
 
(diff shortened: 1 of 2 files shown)


JakeWharton:
Why waste the list and loop instead of just calling the action or setter
directly with the one view?

Orbar:
The intention was to not waste the `index`, but with one view it's meaningless anyway

JakeWharton:
```
butterknife/src/main/java/butterknife/ButterKnife.java:12:8: Unused import - java.util.ArrayList.
butterknife/src/main/java/butterknife/ButterKnife.java:400: Line is longer than 100 characters.
```

JakeWharton:
Squashed and merged. Thanks!

PR #363, opened 2015-09-26
Package ProGuard rules into the aar

### LINKED ISSUE #347: Feature request: Include proguard file in the library
It'd be great to have the proguard configuration included in the library, instead of manually copying all the required changes after each major library update.
This would, of course, require switching to gradle and making the project aar, which also a good move itself.

artem-zinnatullin:
Closes #347!

I'd suggest to update the section about ProGuard on the website right before or after the release to avoid tons of questions "why it does not work" :)

### DIFF
--- a/butterknife-sample/build.gradle
+++ b/butterknife-sample/build.gradle
@@ -17,6 +17,12 @@ android {
     versionCode 1
     versionName '1.0.0'
   }
+
+  buildTypes {
+    debug {
+      minifyEnabled true
+    }
+  }
 }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


JakeWharton:
The website is checked into source control. You can remove the section in this PR.

artem-zinnatullin:
Done!

For some reason, I thought that site will be deployed automatically (never worked with GitHub pages).

JakeWharton:
We deploy the site manually when making a release. This lets it be revisioned with the code changes that affect it.

Issue #358, opened 2015-09-22
BindColor and BindDrawable should retrieve via ContextCompat instead of directly from Resources


wuman:
Starting in M, the `ContextCompat` class provides the `getColor()`, `getColorStateList()` and the `getDrawable()` helper methods to retrieve colors and drawables.  The respective methods in the `Resources` class have been deprecated.

As such we should modify butterknife to accommodate these changes.  This would, however, require butterknife to depend on the support library.  I am happy to provide a patch for this, but I'm unsure whether @JakeWharton is okay with this new dependency.  Let me know!

JakeWharton:
We don't need support-v4, as the ContextCompat methods are just glorified `if` statements that we can replicate in the runtime.

JakeWharton:
This was fixed.

jaredsburrows:
@JakeWharton @wuman What about the `getColor`?

I am seeing plenty of warnings:

```
warning: [deprecation] getColor(int) in Resources has been deprecated
    target.COLOR_PERCENT_CHANGE_RED = res.getColor(2131624030);
```

JakeWharton:
The fix hasn't been released yet.

jaredsburrows:
@JakeWharton Thanks for the quick reply. The date of the PR was merged at the end of September. When do you plan on releasing?

JakeWharton:
I have no plan. Whenever all the things I want to put in 8.0 are done.

Issue #326, opened 2015-08-07
Tint support on @BindDrawable


JakeWharton:
Usage:
```java
@BindDrawable(value = R.drawable.my_sweet_logo, tint = R.attr.accentColor)
Drawable mySweetLogo
```
Generates:
```java
TypedValue outValue = new TypeValue();
context.getTheme().resolveAttribute(annotation.tint(), outValue, true);
Resources res = context.getResources();
Drawable drawable = DrawableCompat.wrap(res.getDrawable(annotation.value()).mutate());
DrawableCompt.setTint(drawable, res.getColor(outValue.resourceId));
target.mySweetLogo = drawable;
```

JakeWharton:
Check for the support-v4's `DrawableCompat` on the classpath as a prerequisite. This allow providing a friendly error message vs. just having the generated code fail to compile.

mattprecious:
The single argument `getDrawable` was deprecated in API 22, so the generated code should do a version check and call the 2-arg version on 22+ and pass a null `Theme` .

JakeWharton:
Ugh, yeah. That's being tracked by #309 because I didn't want to do it in generated code but as a static helper in one place (like the Context helpers we already have).

lukaciko:
Do we really want the user to pass a color attribute instead of color resource ID? The above code would fail at runtime if the user accidentally passed a color resource ID (at `res.getColor()`).

JakeWharton:
It would fail at compile-time with an `@AttrRes` annotation which it will have.

JakeWharton:
Plus that same argument can be made for literally every `int` in this whole library.

lukaciko:
Oh sure, support annotations to the rescue!

Still, don't you think color resources are more convenient in most cases?

JakeWharton:
It's certainly easier to use, but i'm not sure it's more convenient. The
whole point of using a theme attribute is such that it allows the drawables
to change color when the view is inflated in multiple themes. For example,
we have a "green world" theme and a "white world" theme in our app and
views (with drawables) are used in both. We want to reference the same
drawable but have its color reflect the "primary" color of the theme.

JakeWharton:
This was merged.

Issue #309, opened 2015-07-10
@BindDrawable generates deprecated code on API 22+


svenjacobs:
When on API level 22+ the code generated by `@BindDrawable` provokes a deprecation warning because [Resource#getDrawable(int)](https://developer.android.com/reference/android/content/res/Resources.html#getDrawable(int)) should not be used on API 22+ anymore.

JakeWharton:
This only matters if you use theme attributes in a drawable XML.

JakeWharton:
A fix was merged.

PR #364, opened 2015-09-27
Add tint support for @BindDrawable.

lukaciko:
Refs #326, #309 and #358.

I've put drawable tinting code into utility classes to have the option of checking whether support library is on classpath at runtime. Something that I'm unsure of is what to do if `resolveAttribute` returns false (e.g. because the app didn't set the theme with the attribute).


build.gradle
```diff
@@ -45,7 +45,8 @@ ext {
 ext.deps = [
     // Android
     android: 'com.google.android:android:4.1.1.4',
-    
+    supportv4: 'com.android.support:support-v4:23.0.1',
```
JakeWharton: You need to add `extra-android-m2repository` to the Android dependencies in the `.travis.yml`

butterknife-annotations/src/main/java/butterknife/BindDrawable.java
```diff
@@ -16,4 +16,10 @@
 public @interface BindDrawable {
   /** Drawable resource ID to which the field will be bound. */
   int value();
+
+  /** Color attribute resource ID that is used to tint the drawable. */
+  int tint() default NONE;
+
+  /** Default value for {@link #tint()}. */
+  int NONE = -1;
```
JakeWharton: This should be 0, not -1. -1 is technically a valid, possible resource ID. 0 is not a valid resource ID.

Also I'd just inline the default and skip the constant.

butterknife/src/main/java/butterknife/internal/SupportV4Utils.java
```diff
@@ -0,0 +1,42 @@
...
+
+/** Requires v4 Support Library. Check {@link #HAS_SUPPORT_V4_LIBRARY} before use. */
+final class SupportV4Utils {
+  static final boolean HAS_SUPPORT_V4_LIBRARY = hasSupportV4LibraryOnClasspath();
+
+  private SupportV4Utils() {
+    throw new AssertionError("No instances.");
+  }
+
+  static Drawable setDrawableTint(Resources res, Resources.Theme theme, int drawableId,
```
JakeWharton: this should be called `getTintedDrawable`.

butterknife/src/main/java/butterknife/internal/SupportV4Utils.java
```diff
@@ -0,0 +1,42 @@
...
+      int attributeId) {
+    TypedValue outValue = new TypedValue();
```
JakeWharton: This can be a single, static instance. It's wasteful to allocate this instance each time, and since interaction will always happen on the main thread we get thread safety for free.

butterknife/src/main/java/butterknife/internal/SupportV4Utils.java
```diff
@@ -0,0 +1,42 @@
...
+    boolean attributeFound = theme.resolveAttribute(attributeId, outValue, true);
+    if (!attributeFound) {
+      throw new Resources.NotFoundException("Tint attribute not found.");
```
JakeWharton: This exception should include both the attribute ID and do a reverse-lookup on the attribute name. Look at the `findRequiredView` for an example.

butterknife/build.gradle
```diff
@@ -6,6 +6,7 @@ android {
 
   defaultConfig {
     minSdkVersion rootProject.ext.minSdkVersion
+    targetSdkVersion rootProject.ext.targetSdkVersion
```
JakeWharton: Libraries should not have target versions. This can cause problems with manifest merging.

JakeWharton:
This is awesome. Have a few comments, but nothing major.

butterknife/src/main/java/butterknife/internal/SupportV4Utils.java
```diff
@@ -0,0 +1,45 @@
...
+import android.graphics.drawable.Drawable;
+import android.support.v4.graphics.drawable.DrawableCompat;
+import android.util.TypedValue;
+
+import static butterknife.internal.CompatUtils.getColor;
+
+/** Requires v4 Support Library. Check {@link #HAS_SUPPORT_V4_LIBRARY} before use. */
+final class SupportV4Utils {
+  static final boolean HAS_SUPPORT_V4_LIBRARY = hasSupportV4LibraryOnClasspath();
+  static final TypedValue OUT_VALUE = new TypedValue();
```
lukaciko: Should this be all in capital letters, even though it's mutable? Checkstyle complains if it isn't.

JakeWharton: Yeah this is fine. We don't have a strong precedent on mutable "constant" naming anyway.

JakeWharton:
Squashed and merged. Thanks!

PR #366, opened 2015-09-28
Accept vararg Actions in apply calls

### LINKED ISSUE #332: Proposal: Support vararg actions/setter/properties in apply()
Value prop: I've come across cases where I have a few modular actions that are sometimes used at the same time. For the sake of efficiency and convenience, it would be better to be able to apply a variable number of these in a single call and save the unnecessary repeated iterations that result from consecutive applications.

I can add some more concrete examples if necessary, let me know!

hzsweers:
Closes #332

Opted not to do this for properties and setters since it's unlikely that the same value would apply to multiple setters. Open to suggestions on it though

### DIFF
--- a/butterknife/src/main/java/butterknife/ButterKnife.java
+++ b/butterknife/src/main/java/butterknife/ButterKnife.java
@@ -361,10 +361,28 @@ public final class ButterKnife {
     return viewBinder;
   }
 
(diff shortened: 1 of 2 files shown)


JakeWharton:
Makes sense. For setter/property you just have to do multiple calls.

PR #378, opened 2015-10-09
Move unbind method to Unbinder interface.

### LINKED ISSUE #211: Remove listeners in reset
Is there any reason to not remove the listeners as well in the generated `reset()` method? Something like the following:

````
public class ExampleFragment$$ViewInjector {
  public static void inject(Finder finder, final ExampleFragment target, Object source) {
    View view;
    view = finder.findRequiredView(source, 2131493193, "field 'uiList' and method 'onItemClick'");
    target.uiList = (android.widget.ListView) view;
    ((android.widget.AdapterView<?>) view).setOnItemClickListener(
      new android.widget.AdapterView.OnItemClickListener() {
        @Override public void onItemClick(
          android.widget.AdapterView<?> p0,
          android.view.View p1,
          int p2,
          long p3
        ) {
          target.onItemClick(p2);
        }
      });
  }

  public static void reset(ExampleFragment target) {
    // The next line is new!
    target.uiList.setOnItemClickListener(null);
    target.uiList = null;
  }
}
````

### PR DESCRIPTION
**Pull request by serj-lotutovici**
- Split bind and unbind logic in two separate abstractions.
- Null method bindings in generated unbind method(Closes #211).
- Update webpage to reflect the change.

### DIFF
--- a/butterknife-annotations/src/main/java/butterknife/Unbinder.java
+++ b/butterknife-annotations/src/main/java/butterknife/Unbinder.java
@@ -1,0 +1,19 @@
+package butterknife;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+/**
+ * Bind a target field to an un-binder instance. Use when binding in a {@linkplain
+ * android.app.Fragment fragment}.
+ * <pre><code>
+ * {@literal @}Unbinder ButterKnife.Unbinder unbinder;
+ * </code></pre>
+ */
+@Retention(CLASS)
+@Target(FIELD)
+public @interface Unbinder {
+}
--- a/butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
+++ b/butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
@@ -25,11 +25,14 @@ import static butterknife.compiler.ButterKnifeProcessor.NO_ID;
 import static butterknife.compiler.ButterKnifeProcessor.VIEW_TYPE;
 import static java.util.Collections.singletonList;
 import static javax.lang.model.element.Modifier.FINAL;
+import static javax.lang.model.element.Modifier.PRIVATE;
 import static javax.lang.model.element.Modifier.PUBLIC;
+import static javax.lang.model.element.Modifier.STATIC;
 
 final class BindingClass {
   private static final ClassName FINDER = ClassName.get("butterknife.internal", "Finder");
   private static final ClassName VIEW_BINDER = ClassName.get("butterknife.internal", "ViewBinder");
+  private static final ClassName UNBINDER = ClassName.get("butterknife", "ButterKnife", "Unbinder");
   private static final ClassName UTILS = ClassName.get("butterknife.internal", "Utils");
   private static final ClassName VIEW = ClassName.get("android.view", "View");
   private static final ClassName CONTEXT = ClassName.get("android.content", "Context");
@@ -47,11 +50,20 @@ final class BindingClass {
   private final String className;
   private final String targetClass;
   private String parentViewBinder;
+  private UnbinderBinding unbinderBinding;
 
   BindingClass(String classPackage, String className, String targetClass) {
     this.classPackage = classPackage;
     this.className = className;
     this.targetClass = targetClass;
+  }
+
+  void requiresUnbinder(String unbinderFieldName) {
+    unbinderBinding = new UnbinderBinding(classPackage, className, unbinderFieldName);
+  }
+
+  boolean hasRequestedUnbinder() {
+    return unbinderBinding != null;
   }
 
   void addBitmap(FieldBitmapBinding binding) {
@@ -73,8 +85,7 @@ final class BindingClass {
   boolean addMethod(int id, ListenerClass listener, ListenerMethod method,
       MethodViewBinding binding) {
     ViewBindings viewBindings = getOrCreateViewBindings(id);
-    if (viewBindings.hasMethodBinding(listener, method)
-        && !"void".equals(method.returnType())) {
+    if (viewBindings.hasMethodBinding(listener, method) && !"void".equals(method.returnType())) {
       return false;
     }
     viewBindings.addMethodBinding(listener, method, binding);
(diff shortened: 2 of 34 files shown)

### CODE REVIEW

File: butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java [hunk 1]
```diff
@@ -25,11 +25,14 @@
...
 import static java.util.Collections.singletonList;
 import static javax.lang.model.element.Modifier.FINAL;
+import static javax.lang.model.element.Modifier.PRIVATE;
 import static javax.lang.model.element.Modifier.PUBLIC;
+import static javax.lang.model.element.Modifier.STATIC;
 
 final class BindingClass {
   private static final ClassName FINDER = ClassName.get("butterknife.internal", "Finder");
   private static final ClassName VIEW_BINDER = ClassName.get("butterknife.internal", "ViewBinder");
+  private static final ClassName UNBINDER = ClassName.get("butterknife.ButterKnife", "Unbinder");
```
JakeWharton: Should be `ClassName.get("butterknife", "ButterKnife", "Unbinder")`

butterknife-compiler/src/test/java/butterknife/BindUnbinderTest.java
```diff
@@ -0,0 +1,168 @@
...
+        ));
+
+    assertAbout(javaSource()).that(source)
+        .processedWith(new ButterKnifeProcessor())
+        .compilesWithoutError()
+        .and()
+        .generatesSources(expectedSource);
+  }
+
+  @Test public void failOnlyOneUnbinder() {
```
JakeWharton: "failWhenMultipleUnbinders"

butterknife-compiler/src/test/java/butterknife/BindUnbinderTest.java
```diff
@@ -0,0 +1,168 @@
...
+            "}"));
+
+    assertAbout(javaSource()).that(source)
+        .processedWith(new ButterKnifeProcessor())
+        .failsToCompile()
+        .withErrorContaining("Only one filed should be annotated with @BindUnbinder. (test.Test.unbinder2)")
+        .in(source).onLine(7);
+  }
+
+  @Test public void failOnUnbinderTypeAllowed() {
```
JakeWharton: "failOnWrongUnbinderType"

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -122,6 +137,67 @@ JavaFile brewJava() {
...
+    // Only add fields to the unbinder if there are method bindings.
+    Map<ListenerClass, Map<ListenerMethod, Set<MethodViewBinding>>> classMethodBindings =
+        bindings.getMethodBindings();
+    if (classMethodBindings.isEmpty()) {
+      return;
+    }
+
+    // Using view id for name uniqueness.
+    String fieldName = "view" + bindings.getId();
+    result.addField(ClassName.get("android.view", "View"), fieldName);
```
JakeWharton: Use `VIEW` constant

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -122,6 +137,67 @@ JavaFile brewJava() {
...
+
+    // We only need to emit the null check if there are zero required bindings.
+    boolean needsNullChecked = bindings.getRequiredBindings().isEmpty();
+    if (needsNullChecked) {
+      unbindMethod.beginControlFlow("if ($L != null)", fieldName);
```
JakeWharton: Can you include a test that has one of these null checks in it?

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -135,6 +211,11 @@ private MethodSpec createBindMethod() {
       result.addStatement("super.bind(finder, target, source)");
     }
 
+    // If the caller requested an unbinder, we need to create an instance of it.
+    if (unbinderName != null) {
+      result.addStatement("target.$L = new UnbinderImpl()", unbinderName);
```
JakeWharton: Should be `new $T` and pass in `unbinderClassName` from calling method.

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -259,6 +340,12 @@ private void addMethodBindings(MethodSpec.Builder result, ViewBindings bindings)
       result.beginControlFlow("if (view != null)");
     }
 
+    // Add the view reference to the unbinder.
+    if (unbinderName != null) {
+      result.addStatement("((UnbinderImpl) target.$L).$L = view", unbinderName,
```
JakeWharton: Ditto on using `$T` and `unbinderClassName`

JakeWharton: Also not a fan of casting this type every single time. I'd prefer to create it locally, assign all the views, and only set on the `target` at the end.

butterknife-compiler/src/test/java/butterknife/BindUnbinderTest.java
```diff
@@ -0,0 +1,168 @@
...
+            "        target.doStuff();",
+            "      }",
+            "    });",
+            "  }",
+            "  @Override public void unbind(T target) {",
+            "  }",
+            "  private static final class UnbinderImpl implements Unbinder<Test> {",
+            "    View view1;",
+            "    @Override public void unbind(Test target) {",
+            "      view1.setOnClickListener(null);",
```
JakeWharton: The last statement of this method needs to be `target.unbinderFieldName = null`

JakeWharton:
I actually want to get rid of the `ButterKnife.reset` call altogether and force use of the unbinder. That said, let's ignore that until this gets merged and then that can be done as a follow-up.

This looks awesome. I commented all over as there's a few critical things that need changed and a few minor things.

JakeWharton:
I mean `ButterKnife.unbind`... old habits...

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
serj-lotutovici: Done.

serj-lotutovici:
:smile: I know what you mean
> I mean ButterKnife.unbind... old habits...

Seam like an reasonable idea. If you don't mind, I can follow up this such a PR.
And that being said: @JakeWharton what if I remove the newly introduced `@BindUnbinder` and tie the Unbinder binding to `@Bind`?

> I actually want to get rid of the ButterKnife.reset call altogether and force use of the unbinder. That said, let's ignore that until this gets merged and then that can be done as a follow-up.

JakeWharton:
Yeah I'm not a huge fan of `@BindUnbinder`, but I'm not sure how `@Bind` would be re-used as it requires an ID. It could be `@Unbind` or `@Unbinder`.

serj-lotutovici:
Brilliant! I'll change the annotation to `@Unbinder` for now (`@Unbind` sounds more like an action, and I would use that for a method scoped annotation).
> It could be @Unbind or @Unbinder.

If you don't have any other comments regarding the code, I can remove the `#unbind()` method form `ViewBinder` and change the scope of the PR. This way it will be all in one place (and one commit ;)). 

How that does this sound?

JakeWharton:
Just got through reviewing and everything looks good. So if you want to
continue down the path of removing the 'unbind' method please do!

serj-lotutovici:
On it. Don't merge yet!
> Just got through reviewing and everything looks good. So if you want to
> continue down the path of removing the 'unbind' method please do!

JakeWharton:
I would also like to pass the target class instance into the unbinder constructor and remove the argument to the unbind method. This makes it more clear that it's a one-shot deal per instance and doesn't require you passing in 'this' or anything.

serj-lotutovici:
Did I mention how I "love" big pull requests? :smiley:

@JakeWharton Applied all the comments, and requests. Please have a look, when you have time.

butterknife-annotations/src/main/java/butterknife/Unbinder.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+/**
+ * Bind a target field to an un-binder implementation. Recommended to be used when requesting method
+ * bindings in a {@link android.app.Fragment}.
+ * <pre><code>
+ * {@literal @}BindUnbinder ButterKnife.Unbinder unbinder;
```
JakeWharton: should be `Unbinder`

butterknife-annotations/src/main/java/butterknife/Unbinder.java
```diff
@@ -0,0 +1,19 @@
+package butterknife;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+/**
+ * Bind a target field to an un-binder implementation. Recommended to be used when requesting method
```
JakeWharton: "implementation" --> "instance"

JakeWharton: "Recommended to be used when" --> "Use when"

"method bindings in a" --> "binding in a"

"{@link android.app.Fragment}" --> "{@linkplain android.app.Fragment fragment}"

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -114,14 +125,88 @@ JavaFile brewJava() {
...
+    MethodSpec unbinderConstructor = MethodSpec.constructorBuilder()
+        .addParameter(targetClassName, "target")
+        .addStatement("this.$1N = $1N", "target")
+        .build();
+
+    TypeSpec.Builder result =
+        TypeSpec.classBuilder(unbinderBinding.getUnbinderClassName().simpleName())
+        .addSuperinterface(UNBINDER)
+        .addModifiers(PRIVATE, STATIC, FINAL)
+        .addField(targetClassName, "target", PRIVATE)
```
JakeWharton: +`FINAL`

JakeWharton: Ah, nevermind. You null this out.

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -114,14 +125,88 @@ JavaFile brewJava() {
...
+        .addMethod(unbinderConstructor);
+
+    // Even if there are no bindings we need to implement the interface method.
+    MethodSpec.Builder unbindMethod = MethodSpec.methodBuilder("unbind")
```
JakeWharton: The first statement of this method should be:
```java
if (target == null) throw new IllegalStateException("Bindings already cleared.");
```

butterknife-compiler/src/main/java/butterknife/compiler/UnbinderBinding.java
```diff
@@ -0,0 +1,23 @@
+package butterknife.compiler;
+
+import com.squareup.javapoet.ClassName;
+
+class UnbinderBinding {
+  private static final String UNBINDER_IMPL_NAME = "UnbinderImpl";
```
JakeWharton: I don't think we need the "Impl" suffix. Just "Unbinder" will do.

website/index.html
```diff
@@ -161,10 +161,11 @@ <h4 id="listener-binding">Listener Binding</h4>
 </pre>
 
             <h4 id="reset">Binding Reset</h4>
-            <p>Fragments have a different view lifecycle than activities. When binding a fragment in <code>onCreateView</code>, set the views to <code>null</code> in <code>onDestroyView</code>. Butter Knife has an <code>unbind</code> method to do this automatically.</p>
+            <p>Fragments have a different view lifecycle than activities. When binding a fragment in <code>onCreateView</code>, set the views to <code>null</code> in <code>onDestroyView</code>. Butter Knife provides an <code>ButterKnife.Unbinder</code> interface which has an <code>unbind</code> method to do this automatically. Simply bind an unbinder with <code>@Unbinder</code> to the fragment.</p>
             <pre class="prettyprint">public class FancyFragment extends Fragment {
   @Bind(R.id.button1) Button button1;
   @Bind(R.id.button2) Button button2;
+  @Unbinder ButterKnife.Unbinder unbinder;
```
JakeWharton: Hmm.. the only thing I don't like about this is that it prevents you from importing the `Unbinder` interface...

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
serj-lotutovici: I'm not sure if this is necessary. Technically `unbind()` can be called only once. The second time, it will produce a NPE because of `target.unbinder = null`.

JakeWharton: Yes but i'd rather the exception be explicit that it was the user's fault. Otherwise I'll get an issue filed in a few weeks saying that BK cause an NPE.

serj-lotutovici: Fair enough

website/index.html
serj-lotutovici: Yep, thats way I was thinking about using `@Bind`. The idea was to add a default value (like `NO_ID`) to the annotation and check during processing if an `Unbinder` is being bind. But that may brake a lot of stuff.

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
serj-lotutovici: Done

website/index.html
JakeWharton: Let's get this merged and then explore alternatives

JakeWharton:
This is so great. Thanks for all the work!

PR #388, opened 2015-10-10
Add back Optional for methods

### LINKED ISSUE #306: Re-add @Optional for methods only.
Not really an 'issue', just a notice.

The following code will give a lint warning on @Nullable:

**Primitive type members cannot be annotated**

```java
@Nullable
@OnClick(R.id.button1)
public void menuRowClicked(View v) {
    // ...
}
```

hzsweers:
Nullable is problematic on methods since some return primitive types or void, which is both confusing and something IDEs will lint as errors.

Resolves #306

### DIFF
--- a/butterknife-annotations/src/main/java/butterknife/Optional.java
+++ b/butterknife-annotations/src/main/java/butterknife/Optional.java
@@ -1,0 +1,17 @@
+package butterknife;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+/**
+ * Denote that the view specified by the injection is not required to be present.
+ * <pre><code>
+ * {@literal @}Optional @OnClick(R.id.subtitle) void onSubtitleClick;
+ * </code></pre>
+ */
+@Retention(CLASS) @Target(METHOD)
+public @interface Optional {
+}
--- a/butterknife-annotations/src/main/java/butterknife/Optional.java
+++ b/butterknife-annotations/src/main/java/butterknife/Optional.java
@@ -9,7 +9,7 @@ import static java.lang.annotation.RetentionPolicy.CLASS;
 /**
  * Denote that the view specified by the injection is not required to be present.
  * <pre><code>
- * {@literal @}Optional @OnClick(R.id.subtitle) void onSubtitleClick;
+ * {@literal @}Optional @OnClick(R.id.subtitle) void onSubtitleClick() {}
  * </code></pre>
  */
 @Retention(CLASS) @Target(METHOD)
--- a/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
+++ b/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
@@ -875,7 +875,7 @@ public final class ButterKnifeProcessor extends AbstractProcessor {
 
     int[] ids = (int[]) annotationValue.invoke(annotation);
     String name = executableElement.getSimpleName().toString();
-    boolean required = isRequiredBinding(element);
+    boolean required = isOptionalBinding(element);
 
     // Verify that the method and its containing class are accessible via generated code.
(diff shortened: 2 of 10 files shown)


butterknife-annotations/src/main/java/butterknife/Optional.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+/**
+ * Denote that the view specified by the injection is not required to be present.
+ * <pre><code>
+ * {@literal @}Optional @OnClick(R.id.subtitle) void onSubtitleClick;
```
JakeWharton: You didn't quite complete turning this into a method (semicolon)

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1143,17 +1147,19 @@ private String getPackageName(TypeElement type) {
...
+          return true;
+        }
       }
     }
     return false;
   }
 
   private static boolean isRequiredBinding(Element element) {
-    return !hasAnnotationWithName(element, NULLABLE_ANNOTATION_NAME);
+    return !hasAnnotationWithNames(element, NULLABLE_ANNOTATION_NAME, OPTIONAL_ANNOTATION_NAME);
```
JakeWharton: I don't want `@Optional` to work for fields, nor do I want `@Nullable` to work for methods. Each should only work for their respective binding (method vs. field), and only ButterKnife's `@Optional` is supported, not any annotation named "Optional".

JakeWharton:
This mostly looks great, but I want to be a bit more strict in how this annotation is used.

hzsweers:
Oof, just saw the checkstyle fail. Weird that it didn't show up for me when I ran connectedCheck locally. Will fix in a bit along with addressing your comments.

JakeWharton:
I think it only runs for the 'check' task (which, funny enough, does not
run 'connectedCheck')

butterknife-annotations/src/main/java/butterknife/Optional.java
hzsweers: Oops, copy pasta mistake since [the original annotation had it](https://github.com/JakeWharton/butterknife/blob/9970b0a88a9f718f7036a71d86686432caea0f6e/butterknife/src/main/java/butterknife/Optional.java#L13)

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
hzsweers: So should it just do a manual check here as well to make sure they're valid? Or separate `isOptionalBinding` method (and all that comes with that) entirely? 

I was living it to the annotations themselves, but not actually sure if the compiler will yell at you for using, say, a `@Target(METHOD)` annotation on a field or vice versa.

JakeWharton: I thought separate methods thinking there wouldn't be too much else that had to change. If it truly is a lot, I can take a closer look.

hzsweers: I'll give it a shot and see how it goes

hzsweers:
Checkstyle fixed here 2a721d3

butterknife-annotations/src/main/java/butterknife/Optional.java
hzsweers: fed0745

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
hzsweers: Didn't turn out too bad: 9d77edb

JakeWharton:
I have squashed some and split some (like the README), fixed the imports, tweaked some language, and changed `@Optional` to only work with the specified annotation instead of any named "Optional".

Thanks for doing this!

hzsweers:
:+1: thanks for accepting it! Looking forward to 8.0

Issue #390, opened 2015-10-12
@Unbinder usage in BaseFragment


mjurkus:
Up until introduction of `@Unbinder` interface and annotation it was enough to add `Butterknife.unbind(this)` to your `BaseFragment` and all bindings.
From https://github.com/JakeWharton/butterknife/pull/378 it is no longer possible and requires to add
```
@Unbinder
ButterKnife.Unbinder unbinder;
```

to each fragment. Which IMO is not sexy.

```
public class BaseFragment$$ViewBinder<T extends BaseFragment> implements ViewBinder<T> {
  @Override
  public void bind(final Finder finder, final T target, Object source) {
    Unbinder unbinder = new Unbinder(target);
    target.unbinder = unbinder;
  }

  private static final class Unbinder implements butterknife.ButterKnife.Unbinder {
    private BaseFragment target;

    Unbinder(BaseFragment target) {
      this.target = target;
    }

    @Override
    public void unbind() {
      if (target == null) throw new IllegalStateException("Bindings already cleared.");
      target.unbinder = null;
      target = null;
    }
  }
}
```

serj-lotutovici:
This was done, to give the unbinder a reference to all fragment views that have method bindings. See issue #211.  This way we can null out the listeners for the views, that where set via method bindings.

Since the parent fragment, doesn't really know about it's child's views, this approach was taken.

As an extension, of this, there may be a possibility to make it more generic by introducing `parent unbinder` handling, this way `ButterKnife.bind` will use the same field, but assign a different reference in the child. But not sure how this will work out.

JakeWharton:
Sounds like we just need the hierarchy of the generated unbinder to match that of the viewbinder (and calling classes)

serj-lotutovici:
@JakeWharton So basically during annotation processing check if there is a parent unbinder and make the actual unbinder inherit form that? 

But how then we handle the nulling of `target.binder` and `target.unbinder`?

JakeWharton:
Sorry I wasn't thinking it through completely. This might be a bit tricky to get right.

serj-lotutovici:
@JakeWharton I almost got it working... will submit a PR in a few minutes.

JakeWharton:
Awesome!

serj-lotutovici:
Ok, you're right... this is more tricky then I thought. If you're not in a hurry, will try to finish with this tomorrow...

stephanenicolas:
Hi all, 

[...]

PR #392, opened 2015-10-13
Support for unbinder inheritance.

serj-lotutovici:
This address the issue raised in #390. Allows unbinders to inherit form parent `ViewBinder.ViewUnbinder`'s.

__Changes:__
* Renamed `ButterKnife.Unbinder` -> `ButerKnife.ViewUnbinder`.
* By default only the top level Unbinder (no parent case) overrides the `ButterKnife.ViewUnbinder#unbind()` method, making it final. The actual unbindings happen in `generated_source$$ViewBinder.Unbinder#unbind(T)`, this allows all child unbinders to override it not braking the expected behaviour. 
* For every "potential" parent unbinder we generate 2 protected helper methods in the `ViewBinder` class.
* Each child `ViewBInder` overrides at at-least the `createUnbinder` method providing he's "extended" Unbinder. 
* If a child also binds an unbinder, `accessUnbinder` will be overriden, but will be called only from the next child in the hierarchy (if any).

### DIFF
--- a/butterknife-compiler/src/test/java/butterknife/UnbinderTest.java
+++ b/butterknife-compiler/src/test/java/butterknife/UnbinderTest.java
@@ -8,62 +8,78 @@ import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertAbout;
 import static com.google.testing.compile.JavaSourceSubjectFactory.javaSource;
+import static com.google.testing.compile.JavaSourcesSubjectFactory.javaSources;
+import static java.util.Arrays.asList;
 
 public class UnbinderTest {
   @Test public void bindingUnbinder() {
-    JavaFileObject source = JavaFileObjects.forSourceString("test.Test", Joiner.on('\n')
-        .join(
-            "package test;",
-            "import android.support.v4.app.Fragment;",
-            "import butterknife.ButterKnife;",
-            "import butterknife.OnClick;",
-            "import butterknife.Unbinder;",
-            "public class Test extends Fragment {",
-            "  @Unbinder ButterKnife.Unbinder unbinder;",
-            "  @OnClick(1) void doStuff() {",
-            "  }",
-            "}"
-        ));
+    JavaFileObject source = JavaFileObjects.forSourceString("test.Test",
+        Joiner.on('\n')
+            .join(
+                "package test;",
+                "import android.support.v4.app.Fragment;",
+                "import butterknife.ButterKnife;",
+                "import butterknife.OnClick;",
+                "import butterknife.Unbinder;",
+                "public class Test extends Fragment {",
+                "  @Unbinder ButterKnife.ViewUnbinder unbinder;",
+                "  @OnClick(1) void doStuff() {",
+                "  }",
+                "}"
+            ));
 
     JavaFileObject expectedSource = JavaFileObjects.forSourceString("test/Test$$ViewBinder",
-        Joiner.on('\n').join(
-            "package test;",
-            "import android.view.View;",
-            "import butterknife.ButterKnife;",
-            "import butterknife.internal.DebouncingOnClickListener;",
-            "import butterknife.internal.Finder;",
-            "import butterknife.internal.ViewBinder;",
-            "import java.lang.IllegalStateException;",
-            "import java.lang.Object;",
-            "import java.lang.Override;",
-            "public class Test$$ViewBinder<T extends Test> implements ViewBinder<T> {",
-            "  @Override public void bind(final Finder finder, final T target, Object source) {",
-            "    Unbinder unbinder = new Unbinder(target);",
-            "    View view;",
-            "    view = finder.findRequiredView(source, 1, \"method 'doStuff'\");",
-            "    unbinder.view1 = view;",
-            "    view.setOnClickListener(new DebouncingOnClickListener() {",
-            "      @Override public void doClick(View p0) {",
-            "        target.doStuff();",
-            "      }",
-            "    });",
-            "    target.unbinder = unbinder;",
-            "  }",
-            "  private static final class Unbinder implements ButterKnife.Unbinder {",
-            "    private Test target;",
-            "    View view1;",
-            "    Unbinder(Test target) {",
-            "      this.target = target;",
-            "    }",
-            "    @Override public void unbind() {",
-            "      if (target == null) throw new IllegalStateException(\"Bindings already cleared.\");",
-            "      view1.setOnClickListener(null);",
-            "      target.unbinder = null;",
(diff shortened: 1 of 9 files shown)


mjurkus:
@JakeWharton any thoughts on this?

JakeWharton:
Haven't had a chance to look closely yet. Soooooon.

butterknife-compiler/src/test/java/butterknife/UnbinderTest.java
```diff
@@ -190,68 +209,312 @@
...
+                "    view = finder.findRequiredView(source, 1, \"method 'doStuff1'\");",
+                "    unbinder.view1 = view;",
+                "    view.setOnClickListener(new DebouncingOnClickListener() {",
+                "      @Override public void doClick(View p0) {",
+                "        target.doStuff1();",
+                "      }",
+                "    });",
+                "    target.unbinder = unbinder;",
+                "  }",
+                "  protected butterknife.ButterKnife.Unbinder<T> createUnbinder(T target) {",
```
JakeWharton: This could be `<U extends Unbinder<T>> U createUnbinder(T target)` to avoid the cast in calling code

serj-lotutovici: Done, but I couldn't avoid casting in the actual methods (otherwise I was getting a compilation error).

butterknife-compiler/src/test/java/butterknife/UnbinderTest.java
```diff
@@ -190,68 +215,326 @@
...
+                "    unbinder.view1 = view;",
+                "    view.setOnClickListener(new DebouncingOnClickListener() {",
+                "      @Override public void doClick(View p0) {",
+                "        target.doStuff1();",
+                "      }",
+                "    });",
+                "    target.unbinder = unbinder;",
+                "  }",
+                "  @SuppressWarnings(\"unchecked\")",
+                "  protected <U extends butterknife.ButterKnife.Unbinder<T>> U createUnbinder(T target) {",
```
JakeWharton: Sorry I didn't mean `ButterKnife.Unbinder` but the `Unbinder` type that's generated in this class.

serj-lotutovici: Hm... I've changed locally to use the generated `Unbinder` type and now two tests fail with error message: `method does not override or implement a method from a supertype @Override` for the child generated classes. See: `childBindsSecondUnbinder` and `childUsesOwnUnbinder`.

The child classes should drop the generic declaration and use their types instead?

serj-lotutovici: Tried to play with this again, and code wise it doesn't make any difference in relation to the casts, except that in your case every child class has to override a method with `<U extends Parent$$ViewBinder.Unbinder<T>>` (and that including second/third and so on generations) instead of `<U extends ButterKnife.Unbinder<T>>`. And still cast binder to `U` inside the methods. 
Also using the generated `Unbinder` will result in one additional layer of complexity, since, we will have to know the name of the initial parent, this will require an additional loop / or field in the `BindingClass`

serj-lotutovici: @JakeWharton did I misunderstood something regarding your comment? I don't understand why using the generated `Unbinder` class type is better then the actual interface.

mjurkus:
Ping... We are really waiting for this one...

JakeWharton:
Me too

serj-lotutovici:
@JakeWharton please take a look again. 

Accessor methods, now return `<U extends Unbinder<T>>` where the `Unbinder` is the generated unbinder class.

serj-lotutovici:
Ping... Sorry if I'm being annoying, but... Is this pr going anywhere?

JakeWharton:
What's the urgency? I'm busy

mjurkus:
@serj-lotutovici I checked you branch and it seems that there is an issue when parent fragment is in another package. Then from child fragments it cannot be reached. Since that constructor is with default visibility.
```
Unbinder(T target) {
    this.target = target;
}
```

serj-lotutovici:
Oh... I actually didn't think about that... Thx for pointing it out...

mjurkus:
Np. Could you ping if/when you're update it? :)

serj-lotutovici:
@mjurkus pushed a fix for the package visibility issue. I made the constructor for the unbinder `protected` for not (not sure how much sense does it make). Also added a test to check this particular use case (just in case, shouldn't hurt).

mjurkus:
Thanks a lot!

mjurkus:
@serj-lotutovici we noticed another issue.
If we have hierarchy like this
`FragmentFoo` > `SomeAbstractFragment` > `BaseFragment`

and when `BaseFragment` has unbinder injected - view unbinding does not work.

Not a big issue, but just letting you know.

serj-lotutovici:
Can you post the unbinder part of BaseFragment? I was planning to rebase
this commit anyway since Jake updated JavaPoet, might as well take a look
at this...

mjurkus:
Nothing special about it.

```
  @Override
    public void onDestroyView() {
        super.onDestroyView();
        if (unbinder != null) {
            unbinder.unbind();
        } 
    }
```

butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -85,7 +85,7 @@ private ButterKnife() {
 
   /** An unbinder contract that can be bind with {@link butterknife.Unbinder}. */
   @SuppressWarnings("unused") // Used by generated code.
-  public interface Unbinder {
+  public interface ViewUnbinder<T> {
```
serj-lotutovici: Sorry for squeezing this in... but `@Unbinder ButterKnife.Unbinder` is driving me crazy...

serj-lotutovici:
@mjurkus I've tried to reproduce the issue discussed above, and couldn't :( It seams to work form my perspective. May be you can dive me a test that would reproduce the issue?

mjurkus:
I will try to create a sample project with the issue.

mjurkus:
I can't really reproduce it 100%. Sometimes sometimes on different builds there are no unbinder class generated for child fragments. After a clean and recompile - it was generated. Don't really understand why...

For example if we have `LoginFragment` that extends `AbstractAuthFragment` that extends `BaseFragment` login fragment in some cases has LoginFragment$$ViewBinder without unbinder generated. Unbinder was created in `AbstractAuthFragment$$ViewBinder` and `BaseFragment$$ViewBinder`

PR #400, opened 2015-11-05
Explicitly turn off lint for resource bindings

bmeike:
Lint checks resource fetches, now, to be sure that the request type matches the resource type.  It doesn't ignore raw numbers, as it probably should.  This is a workaround that does not require turning of type checking for the whole project.

All test pass.  I have not verified the output in my own test project, because I can't find the build artifact.  If you can tell me where it is, I'll do that.


JakeWharton:
I'd rather add an annotation for this on the method rather than individual comments at the beginning of every single line.

bmeike:
Got it.  Makes sense.  I was trying to turn it off at the smallest affected element.

I also see that you already added the annotation to the method.  I missed that.

Thanks,
  -blake

bmeike:
equivalent fix added previously

JakeWharton:
Oh, cool. I forgot already... Oops!

Issue #438, opened 2016-02-02
BUG


Tinkling:
```java
public class Activity1 extends AppCompatActivity {
    @Bind(R.id.button) Button button1;
}

public class Activity2 extends Activity1 {
    @Bind(R.id.button2) Button button2;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_2);
        ButterKnife.bind(this);
        // ...
    }
}
```
Build project.

Generated classes:
```java
public class Activity1$$ViewBinder<T extends Activity1> implements ViewBinder<T> {
    // ...
}
```
and
```java
public class Activity2$$ViewBinder<T extends Activity2> extends Activity1$$ViewBinder<T> {
    // ...
}
```

Modify Activity2.java(e.g. add a blank line) and build.

Generated classes:
```java
public class Activity1$$ViewBinder<T extends Activity1> implements ViewBinder<T> {
    // ...
}

public class Activity2$$ViewBinder<T extends Activity2> implements ViewBinder<T> {
    // ...
}
```

Activity2$$ViewBinder not extends Activity1$$ViewBinder!

cypressf:
Probably a duplicate of #31

cypressf:
I also see this issue when modifying any unrelated class and rebuilding, not just modifying Activity2.

zoltish:
Seeing this too after doing some work and running the app(s). Seems to always be related to abstract classes doing the Butterknife calls, usually fixed by rebuilding the whole project (although not optimal at all).

Jose-Parente:
See https://code.google.com/p/android/issues/detail?id=200115

Tinkling:
@Jose-Parente Thanks!

Jose-Parente:
I think this should be looked at by the butterknife team... I believe it should work even when java incremental compilation is on. Disabling incremental compilation is a hack, and it will have impact on building time.

Tinkling:
https://code.google.com/p/android/issues/detail?id=200097
bendowski@google.com:"It will be fixed in the next preview release."

PR #439, opened 2016-02-02
Fix:[issue#438](https://github.com/JakeWharton/butterknife/issues/438)

Tinkling:
Fix:[issue#438](https://github.com/JakeWharton/butterknife/issues/438)


Tinkling:
This is not butterknife's issue.

PR #455, opened 2016-03-15
filed -> field in UnbinderTest

egor-n:


serj-lotutovici:
@egor-n You need to update the processor code as well see: https://github.com/JakeWharton/butterknife/blob/master/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java#L775

egor-n:
@serj-lotutovici thanks, I have updated the pull request.

PR #457, opened 2016-03-27
Replace annotations with Resource prefix with Bind prefix

veerenmandalia:
Is this the correct way to update the javadoc for this project?

[deploy_website.sh](https://github.com/JakeWharton/butterknife/blob/master/deploy_website.sh) would overwrite the changes made in this patch unless the javadoc jar is created and uploaded to repository.sonatype.org


JakeWharton:
This was fixed by f6d9fb793cbaa6aff5f574edf1ad6d49c7ad980e and will be included in the next release. I don't really want to be modifying the release Javadoc, but whatever.

PR #459, opened 2016-03-28
Proper/optimized unbinder inheritance and new bind() API

hzsweers:
## Structure

This is two parts folded into one:
- `Unbinder` inheritance is now fully functional and optimized
  - By functional, I mean that the current master implementation didn't always work and failed to compile on the `A-H` test samples I added.
  - By optimized, I mean that generated classes won't generate unnecessary extra methods or `Unbinder` classes unless absolutely necessary, avoiding a lot of generated cruft. Rules around this include:
    - If top level, no view bindings, just return a singleton NOP with no generated `Unbinder` extras
    - If top level, has view bindings, then it generates `Unbinder` extras and returns the real implementation
    - In an inheritance tree, `Unbinder` classes and creators are only generated at the first descendant to need it
    - Future descendants will extend from the highest available `Unbinder` for the `createUnbinder()` parametrization, while extending the most recent `Unbinder` class for its `Unbinder` parent.
    - view-binding-less intermediary descendants won't generate `Unbinder` classes or `createUnbinder()` methods
  - This works by analyzing the tree in the processor. After we find all the top-level classes, we recursively set up the required info in all their descendants to allow `BinderClass` to generate the appropriate code depending on the `Unbinder` situation in the parent(s) it extends.
  - There's a pretty extensive set of test classes in the unbinder package that should cover all the edge cases. Planning to add one more for viewbindings in the base class, though I've manually confirmed this works properly as well.
- All the `bind()` methods now return the corresponding `ViewUnbinder`, rather than injecting the `ViewBinder` on a separate field.
- There is no more `accessUnbinder()` method now since it's no longer necessary.

## Remaining work

- Updating existing tests for the new API and generated code structure. This is just tedium, and I wanted to get this up for code/API review before going through all of it.
- `ViewUnbinder` or go back to `Unbinder` now that there's no annotation?
- `@CheckResult` on `bind()` calls or no?
- User is now responsible for disposing of their `Unbinder` instance themselves if they save it. Their problem, but also might hit friction after how simple `ButterKnife.reset()` was. Thoughts? Do we care?
- You tell me?


hzsweers:
On a side note, this is what trying to sketch out all the edge cases and flow in a coffee shop looks like:

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -247,44 +253,75 @@
...
+        .await();
+
     return targetClassMap;
   }
 
+  /**
+   * Recurse through descendants and set up unbinder inheritance relationships.
+   * This can be multithreaded because there's no multiple inheritance. For small projects, the
+   * difference is negligible. For larger projects though, the overhead of switching schedulers is
+   * amortized. This could especially make a difference for base activities and base fragments.
```
hzsweers: Or it could make none at all. This sounded better to me half awake at 1am ¯\_(ツ)_/¯

butterknife/src/test/java/butterknife/ButterKnifeTest.java
```diff
@@ -230,10 +232,12 @@ public void resetViewsCache() {
   @Test public void bindingViewReturnsView() {
```
JakeWharton: Delete this test.

butterknife/src/main/java/butterknife/internal/Utils.java
```diff
@@ -22,6 +22,8 @@ private static boolean hasSupportV4() {
       return true;
     } catch (ClassNotFoundException ignored) {
       return false;
+    } catch (VerifyError ignored) {
```
JakeWharton: When would this happen?

butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -83,7 +85,7 @@ private ButterKnife() {
     throw new AssertionError("No instances.");
   }
 
-  /** An unbinder contract that can be bind with {@link butterknife.Unbinder}. */
+  /** An unbinder contract that will unbind views when called. */
   @SuppressWarnings("unused") // Used by generated code.
   public interface ViewUnbinder<T> {
```
JakeWharton: Lose the generic parameter? I don't see it serving any value anywhere.

butterknife/src/main/java/butterknife/internal/Utils.java
hzsweers: I dunno, it happened to me locally though which is why I added it. I get the following if I remove it:

```
WARNING: no system properties value for ro.build.date.utc
DEBUG: Loading resources for android from jar:/Users/zsweers/.m2/repository/org/robolectric/android-res/4.1.2_r1_rc/android-res-4.1.2_r1_rc-real.jar!/res...

java.lang.VerifyError: Expecting a stackmap frame at branch target 46
Exception Details:
  Location:
    android/support/v4/graphics/drawable/DrawableCompat.<init>()V @18: ifnull
  Reason:
    Expected stackmap frame at this location.
  Bytecode:
    0x0000000: 2ab7 0093 2ab6 0096 1298 0312 02b8 009e
    0x0000010: 4c2b c600 1c2b 2a2a b600 a203 bd00 04b9
    0x0000020: 00a8 0400 57a7 0014 4d2c b800 acbf 2ab7
    0x0000030: 0090 b14d 2cb8 00ac bfb1               
  Exception Handler Table:
    bci [21, 37] => handler: 40
    bci [46, 50] => handler: 51


	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at butterknife.internal.Utils.hasSupportV4(Utils.java:21)
	at butterknife.internal.Utils.<clinit>(Utils.java:17)
	at com.example.butterknife.unbinder.A$$ViewBinder.bind(A$$ViewBinder.java:21)
	at com.example.butterknife.unbinder.B$$ViewBinder.bind(B$$ViewBinder.java:17)
	at com.example.butterknife.unbinder.C$$ViewBinder.bind(C$$ViewBinder.java:19)
	at com.example.butterknife.unbinder.E$$ViewBinder.bind(E$$ViewBinder.java:17)
	at com.example.butterknife.unbinder.G$$ViewBinder.bind(G$$ViewBinder.java:19)
	at com.example.butterknife.unbinder.H$$ViewBinder.bind(H$$ViewBinder.java:18)
	at com.example.butterknife.unbinder.H$$ViewBinder.bind(H$$ViewBinder.java:14)
	at butterknife.ButterKnife.bind(ButterKnife.java:198)
	at butterknife.ButterKnife.bind(ButterKnife.java:178)
	at com.example.butterknife.unbinder.A.<init>(A.java:14)
	at com.example.butterknife.unbinder.B.<init>(B.java:14)
	at com.example.butterknife.unbinder.C.<init>(C.java:16)
	at com.example.butterknife.unbinder.E.<init>(E.java:14)
	at com.example.butterknife.unbinder.G.<init>(G.java:17)
	at com.example.butterknife.unbinder.H.<init>(H.java:16)
	at com.example.butterknife.unbinder.UnbinderTest.verifyContentViewBinding(UnbinderTest.java:32)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.robolectric.RobolectricTestRunner$2.evaluate(RobolectricTestRunner.java:234)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.robolectric.RobolectricTestRunner$1.evaluate(RobolectricTestRunner.java:175)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:234)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:74)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:497)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)


Process finished with exit code 255
```

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -256,57 +259,38 @@ private void addFieldAndUnbindStatement(TypeSpec.Builder result, MethodSpec.Buil
...
-        accessMethod.addAnnotation(Override.class);
-      }
-      viewBindingClass.addMethod(accessMethod.build());
-    }
   }
 
   private MethodSpec createBindMethod() {
     MethodSpec.Builder result = MethodSpec.methodBuilder("bind")
         .addAnnotation(Override.class)
+        .returns(UNBINDER)
```
JakeWharton: nit: move down one line. I like these to read like the method does:
```
@Override public ViewUnbinder
```

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -393,10 +381,19 @@ private MethodSpec createBindMethod() {
       }
     }
 
+    // Finally return the unbinder
+    if (hasParentBinding() || hasViewBindings()) {
+      result.addStatement("return unbinder");
+    } else {
+      result.addStatement("return butterknife.ButterKnife.NOP_VIEW_UNBINDER");
```
JakeWharton:
```
result.addStatement("return $T.NOP_VIEW_UNBINDER", ButterKnife.class);
```

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -247,44 +253,75 @@
...
-      if (!SuperficialValidation.validateElement(element)) continue;
-      try {
-        parseBindUnbinder(element, targetClassMap, erasedTargetNames);
-      } catch (Exception e) {
-        logParsingError(element, Unbinder.class, e);
-      }
-    }
-
     // Try to find a parent binder for each.
+    HashSet<BindingClass> topLevelClasses = new HashSet<>();
```
JakeWharton: nit: `Set` on left, `LinkedHashSet` on right.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -247,44 +253,75 @@
...
     }
 
+    Observable.from(topLevelClasses)
+        .flatMap(topLevelClass -> {
+          if (topLevelClass.hasViewBindings()) {
+            // It has an unbinder class and it will also be the highest unbinder class for all
+            // descendants
+            topLevelClass.setHighestUnbinderClassName(topLevelClass.getUnbinderClassName());
+          } else {
+            // No unbinder class. Null it out so we know we can just return the NOP unbinder
```
JakeWharton: All your comments are missing periods. I'm commenting on the most problematic one because it contains a period for the first sentence but not the second. [Comments should be sentences](http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html).

butterknife/src/main/java/butterknife/ButterKnife.java
hzsweers: I had the same thought. I'm not totally familiar with the original reasoning for it, but I think it's unfortunately necessary. In particular, this seems to be necessary in order to allow proper overriding of the `void unbind(T target)` method in subclassed `Unbinder` classes. Otherwise, if you have type `C$$ViewBinder.Unbinder`, and later have `G$$ViewBinder.Unbinder extends C$$ViewBinder.Unbinder`, then the subclass will still expect a type `C` in its `unbind()` method. Not sure if there's some generics-fu I'm missing here that could solve this.

butterknife-sample/src/test/java/com/example/butterknife/unbinder/UnbinderTest.java
```diff
@@ -0,0 +1,52 @@
...
+    H h = new H(frameLayout);
+
+    ButterKnife.ViewUnbinder unbinder = ButterKnife.bind(h, frameLayout);
+    verifyHBound(h);
+    unbinder.unbind();
+    verifyHUnbound(h);
+  }
+
+  private void verifyHBound(H h) {
+    Truth.assertThat(h.button1).isNotNull();
```
JakeWharton: static import

butterknife/src/test/java/butterknife/ButterKnifeTest.java
hzsweers: 04abeddb7ecca33c18504522a6b8b2fc84d784e9

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
hzsweers: c35aa8f213a33d0213e97444f916603c2c3fbe9e

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
hzsweers: ButterKnife isn't on the classpath of the compiler module. Can do a `ClassName` representation of it though

6f7b05819d73b682397ba891c367a326ed882b0d

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
hzsweers: c3c63e0929eeeb56d1282e6fb2c026b659804beb

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
hzsweers: Ah, my bad. Didn't realize there was a convention on inline comments.

a3023e623040e832c10d58a9fbae86482f843212

butterknife-sample/src/test/java/com/example/butterknife/unbinder/UnbinderTest.java
hzsweers: 7014cccd87af1a1c9f02ab1f9eb506e16306c9b6

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -39,31 +42,40 @@
   private static final ClassName CONTEXT = ClassName.get("android.content", "Context");
   private static final ClassName RESOURCES = ClassName.get("android.content.res", "Resources");
   private static final ClassName THEME = RESOURCES.nestedClass("Theme");
+  private static final ClassName BUTTERKNIFE = ClassName.get("butterknife", "ButterKnife");
   private static final ClassName UNBINDER =
```
ChristianBecker: This can be `BUTTERKNIFE.nestedClass("ViewUnbinder")` to be in-line with `RESOURCES` and `THEME`

hzsweers: Neat. e6ecd67d123424fb543437444956690fedbdb7d5

butterknife/src/main/java/butterknife/ButterKnife.java
JakeWharton: The `unbind` method takes no argument. And method parameter types are not covariant so you wouldn't be able to override with a more specific argument type anyway.

butterknife-sample/src/main/java/com/example/butterknife/unbinder/H.java
```diff
@@ -0,0 +1,19 @@
...
+public class H extends G {
+
+    @BindColor(android.R.color.primary_text_dark) @ColorInt int grayColor;
+    @Bind(android.R.id.button3) View button3;
+
+    public H(View view) {
+        super(view);
+        ButterKnife.bind(this, view);
+    }
+}
```
JakeWharton: All these new classes should be two-space indent

JakeWharton:
> @CheckResult on bind() calls or no?

No. Don't want to force people to care about unbinding when they don't have to (e.g. a view binding to its children).

JakeWharton:
> User is now responsible for disposing of their Unbinder instance themselves if they save it. Their problem, but also might hit friction after how simple ButterKnife.reset() was. Thoughts? Do we care?

Don't care.

JakeWharton:
> ViewUnbinder or go back to Unbinder now that there's no annotation?

`ButterSheath`?

No `Unbinder` sounds good.

JakeWharton:
> Updating existing tests for the new API and generated code structure. This is just tedium, and I wanted to get this up for code/API review before going through all of it.

My body is ready.

butterknife-sample/src/main/java/com/example/butterknife/unbinder/H.java
hzsweers: aface0d

butterknife/src/main/java/butterknife/ButterKnife.java
hzsweers: Ah you're right. I was stupidly trying to remove all parametrization rather than just the initial `implements` form. The inner unbinders still need to be parametrized, but the original implements does not. Fixed in c4f4aaa

hzsweers:
Renamed to Unbinder - d3491d4
Moved Unbinder to top level, renamed generated unbinders to UnnerUnbinder - 30752b7
  - This makes Unbinder a first class citizen since people will be receiving instances of it, and also distinguishes generated inner classes to better differentiate.

Will work on updating the other tests now

hzsweers:
Tests updated, only remaining thing is whether or not you want to support jdk 7 for the compiler (didn't think about that while writing the RxJava stuff). I can change everything to anonymous inner classes if so

JakeWharton:
I would prefer to not force JDK 8. People will be complaining enough about the API change as-is so best not to force yet another thing on them.

hzsweers:
Done. Side note: there were many times while writing this that I wished I were doing it in kotlin. Kotlin for annotation processing seems like a great fit

JakeWharton:
That can be a follow-up :grinning: . I totally agree.

butterknife-compiler/src/test/java/butterknife/BindArrayTest.java
```diff
@@ -20,24 +23,25 @@
...
+        + "import java.lang.Object;\n"
+        + "import java.lang.Override;\n"
+        + "import java.lang.SuppressWarnings;\n"
+        + "public class Test$$ViewBinder<T extends Test> implements ViewBinder<T> {\n"
+        + "  @Override\n"
+        + "  @SuppressWarnings(\"ResourceType\")\n"
+        + "  public Unbinder bind(final Finder finder, final T target, Object source) {\n"
+        + "    Resources res = finder.getContext(source).getResources();\n"
+        + "    target.one = res.getStringArray(1);\n"
+        + "    return ButterKnife.NOP_UNBINDER;\n"
```
JakeWharton: Let's put this implementation on `Unbinder` itself and call it `EMPTY`

butterknife-compiler/src/test/java/butterknife/BindTest.java
```diff
@@ -22,22 +25,45 @@
...
+        + "    View view;\n"
+        + "    view = finder.findRequiredView(source, 1, \"field 'thing'\");\n"
+        + "    target.thing = view;\n"
+        + "    return unbinder;\n"
+        + "  }\n"
+        + "  @SuppressWarnings(\"unchecked\")\n"
+        + "  protected <U extends InnerUnbinder<T>> U createUnbinder(T target) {\n"
+        + "    return (U) new InnerUnbinder(target);\n"
+        + "  }\n"
+        + "  public static class InnerUnbinder<T extends Test> implements Unbinder {\n"
```
JakeWharton: This class can be `protected` too, right? Only subclasses of `Test$$ViewBinder` will have nested classes which subclass.

butterknife-compiler/src/test/java/butterknife/BindArrayTest.java
hzsweers: ooo good idea. Done in b8e7ddaad6b904b425f934c1dc6544f9ba246ff4

butterknife-compiler/src/test/java/butterknife/BindTest.java
```diff
@@ -111,22 +160,45 @@
...
+        + "  @Override\n"
+        + "  public Unbinder bind(final Finder finder, final T target, Object source) {\n"
+        + "    InnerUnbinder unbinder = createUnbinder(target);\n"
+        + "    View view;\n"
+        + "    view = finder.findRequiredView(source, 1, \"field 'thing'\");\n"
+        + "    target.thing = finder.castView(view, 1, \"field 'thing'\");\n"
+        + "    return unbinder;\n"
+        + "  }\n"
+        + "  @SuppressWarnings(\"unchecked\")\n"
+        + "  protected <U extends InnerUnbinder<T>> U createUnbinder(T target) {\n"
```
JakeWharton: No need for the type variable, casting, or suppression. Just return `InnerUnbinder<T>` here and in subclasses.

Example:
```java
class View {}
class TextView extends View {}

class A<T extends View> {
  protected Inner<T> inner() {
    return new Inner<>();
  }
  
  protected static class Inner<T> {
    
  }
}

class B<T extends TextView> extends A<T> {
  @Override protected Inner<T> inner() {
    return new Inner<>();
  }

  protected static class Inner<T> extends A.Inner<T> {
    
  }
}
```

butterknife-compiler/src/test/java/butterknife/BindTest.java
hzsweers: yup, good catch. fa71bd8

JakeWharton:
I have some other small things I'd like to play with here, but I think we should get this merged before doing anything else.

JakeWharton:
Plus it'll give a chance to place with snapshots in real projects to influence decisions.

butterknife-compiler/src/test/java/butterknife/BindTest.java
hzsweers: Done f688855

hzsweers:
Sounds good, and curious to hear what the other things you'd want to play with are

PR #484, opened 2016-04-23
Fix proguard warning

hamberluo:
Fix proguard warning of butterknife.internal.ButterKnifeProcessor


JakeWharton:
This class is no longer in the main artifact on master so ProGuard never sees it.

PR #487, opened 2016-04-23
Latest RxJava

JakeWharton:


build.gradle
```diff
@@ -54,7 +54,7 @@ ext.deps = [
     javapoet: 'com.squareup:javapoet:1.6.1',
 
     // Other
-    rxjava: 'io.reactivex:rxjava:1.1.2',
+    rxjava: 'io.reactivex:rxjava:1.1.4',
```
hzsweers: Am I missing something or shouldn't this be 1.1.3?

JakeWharton: Hello I am from the future!

PR #495, opened 2016-04-24
Portuguese Translation

WilderPereira:
A lot of Brazilians use Butter Knife, so we've translated the website to portuguese.


JakeWharton:
Unfortunately I have no way of maintaining this as the project moves forward. If you want to write a guide in Portuguese (such as in blog form) that works a lot better. Thanks for taking the time to do this though.

PR #499, opened 2016-04-24
Make newlines consistent in compiler unit tests.

JayNewstrom:
This PR makes use of `Joiner.on('\n').join` rather than adding a `\n` to the end of every line in tests.
There were a few tests that used both methods, now they all use the same.

I noticed that checkstyle must not be taking the test source set into account.


JakeWharton:
I prefer the existing way since it's much easier to edit and supports copy/paste.

PR #500, opened 2016-04-24
Enable checkstyle for test/androidTest source sets.

JayNewstrom:
This build will fail, I can assist in cleaning up the checksyle errors if you want this merged.


JakeWharton:
I don't think this is worthwhile. I don't want to deal with things like the column limit since we embed generated strings.

PR #512, opened 2016-04-27
Fix built-in proguard rules for preventing all names from being obfuscated

### LINKED ISSUE #509: Built-in ProGuard rules (ver. 8.0.0) prevents all names from being obfuscated
It looks like exactly the same problem as described in https://github.com/JakeWharton/butterknife/issues/104

Probably because you are using
`-keepnames class *`
instead of
`-keepclasseswithmembernames class *`

Bye.

shiraji:
Close #509

### DIFF
--- a/butterknife/proguard-rules.txt
+++ b/butterknife/proguard-rules.txt
@@ -3,4 +3,4 @@
 
 # Prevent obfuscation of types which use ButterKnife annotations since the simple name
 # is used to reflectively look up the generated ViewBinder.
--keepnames class * { @butterknife.* *; }
+-keepclasseswithmembernames class * { @butterknife.* *; }


JakeWharton:
As you can see by the failing build this doesn't actually work. Classes are still obfuscated which contain BK annotations on their fields and methods.

shiraji:
OK, and the following rule doesn't work, as well

```
-keepclasseswithmembernames class * {
    @butterknife.* <fields>;
}

-keepclasseswithmembernames class * {
    @butterknife.* <methods>;
}
```

I should have researched ProGuard more before sending pull request.

JakeWharton:
Yeah I tried that too. Not sure what's needed here to get the desired
beahvior.

shiraji:
Thanks for taking your time. Let me close this pull request, since I don't have any other options for this issue.
 
Now I hate ProGuard...

PR #535, opened 2016-05-06
Allow listener annotations to specify remove method.

### LINKED ISSUE #521: Unbind TextWatcher
Unbind should call `removeTextChangedListener(listener)` instead of `addTextChangedListener(null)`.

https://github.com/JakeWharton/butterknife/blob/1bc7bc1dab9c633afd4dbd49c5cc18cb16043bd7/butterknife-compiler/src/test/java/butterknife/OnTextChangedTest.java#L76

serj-lotutovici:
By specifying a remover method listener classes can ensure a proper
unbinding behaviour. If not specified the compiler will use the setter.
This also fixes the issue with `TextWatcher` unbinding.

Closes #521


serj-lotutovici:
Just saw, why this wrong... will update the pr.

serj-lotutovici:
So... the logic goes as follows:
If the `ListenerAnnotation` does not require a `remover` method the compiler will follow the standard logic:
  * create the listener as an anonymous class and directly set it via the setter
  * during unbinding call the setter method and set the listener to `null`

However, if it does require the listener to be removed, some magic will happen:
  * the unbinder will hold a strong reference to listener
  * the listener will be unbinded via the provided `remove` method
  * the listener will be nulled as well.

Please correct me if I miss anything, or if this solution is even feasible.

JakeWharton:
Sounds correct. I'll look tomorrow.

PR #550, opened 2016-05-07
Finder is needed in method binds to cast to the desired type

JayNewstrom:
All tests were passing before, but the generated code would fail to compile when the finder was used within a method binding.


butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -350,8 +350,12 @@ private MethodSpec createNewBindToTargetMethod() {
     }
 
     if (bindNeedsFinder()) {
-      result.addParameter(FINDER, "finder")
-          .addParameter(Object.class, "source");
+      if (hasMethodBindings()) {
```
JakeWharton: Can we make this more exact so that it's only `final` when actually needed inside the anonymous class?

JayNewstrom: Sure, I'll look into it.

JakeWharton: Actually don't worry about it. I'm making other changes to generated code and I'll rebase your change on top of it (and add this extra functionality) once I'm done.

JayNewstrom: Sounds good.

JayNewstrom:
I rebased, and added the more specific check for adding the final modifier to the finder.

butterknife-compiler/src/main/java/butterknife/compiler/BindingClass.java
```diff
@@ -730,6 +734,24 @@ private boolean hasMethodBindings() {
...
   }
 
+  private boolean methodBindingsNeedFinder() {
+    for (ViewBindings viewBindings : viewIdMap.values()) {
+      for (Map<ListenerMethod, Set<MethodViewBinding>> listenerMethodSetMap :
+          viewBindings.getMethodBindings().values()) {
+        for (Set<MethodViewBinding> methodViewBindings : listenerMethodSetMap.values()) {
+          for (MethodViewBinding methodViewBinding : methodViewBindings) {
+            for (Parameter parameter : methodViewBinding.getParameters()) {
+              if (!parameter.getType().toString().equals(VIEW.toString())) {
```
JakeWharton: This should be using `parameter.requiresCast()` which is the logic the original method uses.

JayNewstrom: Yeah, I probably should have looked at what that was doing. Saw you pushed a commit here: https://github.com/JakeWharton/butterknife/commit/f8d673d1724a330294a69d0325d6ca2c075a7edf feel free to just take that instead.

JakeWharton: I delete the branch after you updated yours! You can update if you want, otherwise I'll get this merged manually soon with the update.

JayNewstrom: Ok, I'll update this quick.

JakeWharton:
Thanks!

PR #569, opened 2016-05-11
Replace old annotation

nikitin-da:


JakeWharton:
Thanks. I'll deploy this shortly.

PR #570, opened 2016-05-11
Remove unused variable.

JayNewstrom:


JakeWharton:
I saw this about 10 times while making other changes and never circled back to remove it. Thanks

PR #573, opened 2016-05-13
ButterKnife extensions

JayNewstrom:
I'm sending this up as an early review. Please take a look and let me know if this is the direction you were thinking.

What's left:
- [ ] Documentation on `ButterKnifeExtension`
- [ ] `ListenerClass`/`ListenerMethod` moved to a non internal package
- [ ] Failure case testing (what if a non `@ListenerClass` is passed in, etc)
- [ ] Lint is crying wolf in the IDE about the call to ServiceLoader.load (command line is happy though)

### DIFF
--- a/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeExtension.java
+++ b/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeExtension.java
@@ -1,0 +1,9 @@
+package butterknife.compiler;
+
+import java.lang.annotation.Annotation;
+import java.util.List;
+
+public abstract class ButterKnifeExtension {
+
+  public abstract List<Class<? extends Annotation>> listenerClasses();
+}
--- a/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
+++ b/butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
@@ -48,6 +48,7 @@ import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.Set;
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.Filer;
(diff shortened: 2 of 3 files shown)


butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeExtension.java
```diff
@@ -0,0 +1,9 @@
+package butterknife.compiler;
```
ChristianBecker: I think this should be refactored to the `butterknife` or `butterknife-annotations` artifact so that it's on the classpath while writing an app?

JayNewstrom: I'm fairly sure this isn't possible because the classes are needed at compile time by the annotation processor.

In order to use this in an app, I think you'd have to add two extra modules.

1. Runtime custom view you want a butterknife bound listener (view code, and listener class annotation)
2. Compile time subclass of `ButterKnifeExtension`

GrahamRogers:
I've been looking into adding library support for `BindView` (see #605) and saw this pull request. Since there seems to be a lot of controversy over whether library projects should be supported, an extension would be the perfect way to implement it without being endorsed by the ButterKnife team.

However, this implementation of extensions would not work for that use case since it only supports listeners. I would therefore like to propose something like this:

```java
public abstract class ButterKnifeExtension {

  public abstract List<Class<? extends Annotation>> annotationClasses();

  public abstract ??? parseAnnotation(???);
}
```

(I haven't fully thought through what the signature of `parseAnnotation` would be.)

Listener extensions can be supported by having another class like so:

```java
public abstract class ButterKnifeListenerExtension extends ButterKnifeExtension {

  @Override
  public ??? parseAnnotation(???) {
    // Something resembling findAndParseListener
  }
}
```

`parseAnnotation` would replace the call to `findAndParseListener` in `ButterKnifeProcessor`. It should then be possible to create a `BindLibraryView` annotation and a `BindLibraryViewExtension` for my use case. It might even make sense to create a `ButterKnifeFieldExtension` class for binding to fields, but more investigation would be needed. In my opinion, it should be possible to implement every ButterKnife annotation as an extension (with the standard ones enabled by default of course).

If the ButterKnife team think this is a good idea I'd be happy to investigate more, or @JayNewstrom can adapt this pull request.

JayNewstrom:
@GrahamRogers I like your idea, and think it's a good long term plan. However, this PR is as basic as it gets, and is intended to just get the ball rolling with extension support. I'd be happy to continue to contribute in this area once the direction is chosen.

JakeWharton:
Can you rebase this? I'm supportive of the addition.

JayNewstrom:
Rebased!

JakeWharton:
I'm going to defer this until the next release since I want to give it
proper attention before committing to anything.

trevjonez:
Any roadmap for this? Currently wishing butterknife would bind me directly to RxBindings and this seems like the best path towards that.

PR #580, opened 2016-05-17
format code

mlick:
format SimpleActivity code


JakeWharton:
The existing format is better.

Issue #579, opened 2016-05-16
Java Compiler Lint: Avoid warnings?


rfgamaral:
> :app:compileFlavorDebugJavaWithJavac
> Note: C:\Workspace\Project\app\build\generated\source\apt\flavor\debug\com\company\appname\ui\debug\DebugActivity$$ViewBinder.java uses unchecked or unsafe operations.
> Note: Recompile with -Xlint:unchecked for details.

Anyway to disable these Lint checks **just** for generated files (inside the `build` folder only) or at least for ButterKnife?

JakeWharton:
We can generate suppressions. What's the offending code look like?

rfgamaral:
This is the generated class:

```
import android.view.View;
import butterknife.Unbinder;
import butterknife.internal.Finder;
import butterknife.internal.ViewBinder;
import java.lang.IllegalStateException;
import java.lang.Object;
import java.lang.Override;

public class DebugActivity$$ViewBinder<T extends DebugActivity> implements ViewBinder<T> {
  @Override
  public Unbinder bind(final Finder finder, final T target, Object source) {
    InnerUnbinder unbinder = createUnbinder(target);
    View view;
    view = finder.findRequiredView(source, 2131624041, "field 'mDrawerLayout'");
    target.mDrawerLayout = finder.castView(view, 2131624041, "field 'mDrawerLayout'");
    return unbinder;
  }

  protected InnerUnbinder<T> createUnbinder(T target) {
    return new InnerUnbinder(target);
  }

  protected static class InnerUnbinder<T extends DebugActivity> implements Unbinder {
    private T target;

    protected InnerUnbinder(T target) {
      this.target = target;
    }

    @Override
    public final void unbind() {
      if (target == null) throw new IllegalStateException("Bindings already cleared.");
      unbind(target);
      target = null;
    }

    protected void unbind(T target) {
      target.mDrawerLayout = null;
    }
  }
}
```

And these are the full lint errors:
> \app\build\generated\source\apt\variant\debug\com\package\ui\debug\DebugActivity$$ViewBinder.java:23: warning: [unchecked] unchecked call to InnerUnbinder(T) as a member of the raw type InnerUnbinder
>     return new InnerUnbinder(target);
>            ^
>   where T is a type-variable:
>     T extends DebugActivity declared in class InnerUnbinder
[...]

PR #582, opened 2016-05-18
Add diamond operator to non-final InnerUnbinder

ChristianBecker:
I ran into many warnings using the latest ButterKnife version:
```
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```

Recompiling the code with `-Xlint:unchecked` revealed the source of the warnings:
```
[unchecked] unchecked call to InnerUnbinder(T,Finder,Object,Resources,Resources.Theme) as a member of the raw type InnerUnbinder
    return new InnerUnbinder(target, finder, source, res, theme);
           ^
  where T is a type-variable:
    T extends FloatingActionsMenu declared in class InnerUnbinder
```

I fixed this warning by using the diamond operator for non-final `InnerUnbinder` classes. 

Closes #579.

PR #586, opened 2016-05-21
Suppress warnings

jaredsburrows:
@JakeWharton @rfgamaral Fix for https://github.com/JakeWharton/butterknife/issues/579


JakeWharton:
#582 is the right fix here. There's no need to suppress.

PR #591, opened 2016-05-24
Allow using non-final IDs with @BindView.

GrahamRogers:
Implementation of allowing use of non-final resource IDs produced by Android library projects. Backwards compatible with the current method (and I would say it should remain the best practice for non-library projects).

In #2, you say you don't want to support this case unless someone puts forward a good use case. At Google, we have our own build system which we use for (almost) every project (open sourced as Bazel http://bazel.io). Part of the best-practices for using this build system is to have a separate library for each Java package. This means every Java package in an Android app built using this tool at Google is a separate Android library. We also put package-relevant resources with the package, i.e. each package has its own res/ directory. This means all of our resources are part of Android libraries and hence none of the IDs are final. So this is a requirement for any Google app to use Butter Knife.

This pull request only fixes the problem for BindView, but it should be possible to apply a similar pattern to other bindings. One design question I came across was whether it would make sense to create a new annotation, say `@Id` and then have `@BindView(id = @Id("my_view"))` or `@BindView(id = @Id("my_view", res = other.package.R.id.class))`. I think `@Id` might be easier to extend to other bindings but potentially too boilerplatey.

Please take a look and let me know what you think.

### DIFF
--- a/butterknife-annotations/src/main/java/butterknife/BindView.java
+++ b/butterknife-annotations/src/main/java/butterknife/BindView.java
@@ -13,9 +13,28 @@ import static java.lang.annotation.RetentionPolicy.CLASS;
  * <pre><code>
  * {@literal @}BindView(R.id.title) TextView title;
  * </code></pre>
+ *
+ * If used in a library project where the fields are not final, use the other attributes:
+ * <pre><code>
+ * {@literal @}BindView(idName = "title") TextView title;
+ * </code></pre>
  */
 @Retention(CLASS) @Target(FIELD)
 public @interface BindView {
-  /** View ID to which the field will be bound. */
-  @IdRes int value();
+  /** View ID to which the field will be bound. If used, this should be specified on it's own. */
+  @IdRes int value() default -1;
+
+  /**
+   * The name of the ID to use. For example, if the generated ID is referenced by
+   * {@code R.id.my_view} then this should be set to {@code "my_view"}.
+   *
+   * @see #idClass() to specify a generated resource class in a different package.
+   */
+  String idName() default "";
+
+  /**
+   * Generated {@code R.id} class which should be used with {@link #idName}. If not specified, it is
+   * assumed there is one available in the package this annotation is used in.
+   */
+  Class<?> idClass() default void.class;
 }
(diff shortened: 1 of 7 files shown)


GrahamRogers:
Previous solutions to this use getIdentifier to get the non-final IDs. This solution references the generated R class directly, giving compile-time safety. See the `BindViewTest`.

tpfaff:
Nice, excited to see what comes of this.

butterknife-annotations/src/main/java/butterknife/BindView.java
```diff
@@ -13,9 +13,28 @@
  * <pre><code>
  * {@literal @}BindView(R.id.title) TextView title;
  * </code></pre>
+ *
+ * If used in a library project where the fields are not final, use the other attributes:
+ * <pre><code>
+ * {@literal @}BindView(idName = "title") TextView title;
```
kageiit: Is this approach IDE refactor friendly?

oguzbabaoglu:
I tried a similar thing here: https://github.com/oguzbabaoglu/butterfork

One improvement I had in mind was to use an IDE plugin instead of a Gradle plugin to generate the reference class (B). Maybe such an IDE plugin can be considered for this PR too.

Possible benefits of using an IDE plugin and a generated reference class:

- Less error prone than typing out a String
- Support for IDE refactors and autocomplete
- Reference class can stay in sync with R without need to recompile
- Links to the layout or res file just like R references

butterknife-annotations/src/main/java/butterknife/BindView.java
GrahamRogers: Android Studio sometimes refactors names if it finds them in strings, I haven't tried with this method but that might apply. However, if this string isn't changed then assuming you don't have anything else with the same ID (which I think would be changed with an IDE refactor anyway?) then this will cause a compile-time error as the ID will no longer exist.

GrahamRogers:
Our build system has to be able to run without an IDE, so an IDE plugin to generate code won't work for us. A plugin which treats the string like an actual reference to the ID would be cool though.

butterknife-annotations/src/main/java/butterknife/BindView.java
kageiit: What if one were to rename two ids to each other?

GrahamRogers: That can cause an issue, but I think this is the closest you can get to being safe without being able to reference the ID directly. I would imagine swapping the IDs of two views would be rare enough that it won't cause many issues.

kageiit:
@GrahamRogers See #613

GrahamRogers:
Closing since this is provided by #613.

PR #613, opened 2016-06-11
Support Android Library Projects

kageiit:
## Summary
There have been many solutions proposed to support using Butterknife in Android Libraries i.e non-final IDs in R. This approach uses a combination of Gradle Plugin + Annotation Processing + Custom Lint to ensure this is done as safely as possible with minimal configuration.

## Usage
- Apply gradle plugin to the library project. 
```
apply plugin: 'butterknife'
```
The plugin generates an `R2` class which is essentially a copy of `R` with final values and the correct support annotations.

- Specify the `R` class for the library using the `@RClass` annotation, just once, anywhere in code. Use `R2` instead of R in all butterknife annotations. 
```java
package butterknife.library;

import android.view.View;
import butterknife.BindView;
import butterknife.OnClick;
import butterknife.RClass;

@RClass(R.class)
class Foo {

  @BindView(R2.id.footer) View bar;

  @OnClick(R2.id.hello) void sayHello() { }
}
```
The `@Rclass` annotation is used by the annotation processor to map the integer ids to symbols in `R`. This in turn generates code like
```java
target.bar = finder.findRequiredView(source, com.example.butterknife.library.R.id.footer, "field 'bar'");
view = finder.findRequiredView(source, com.example.butterknife.library.R.id.hello, "method 'sayHello'");
```
in the `ViewBinder`. This ensures that the final values used from `R` values will be generated by the app consuming the library.
- If `R2` is being used outside butterknife annotations, a custom lint rule will throw an error
screen shot 2016-06-10 at 9 54 12 pm
This is important because values in `R2` are final and would be inlined as constants when used outside annotations.
- Profit!

## Pros
- Api invisible to consumers of the library. `R2` is not packaged into the aar just like `R`.
- Full compile time safety. No referring to `R` values using strings.
- Migrating existing library projects to take advantage of this approach is very simple.
- No changes to the butterknife annotation APIs.
- Backward compatible with existing app projects that use butterknife. In fact, just adding `@RClass` will output human friendly generated code in existing app projects.
- Forward compatible with Butterknife extensions (#573) as the approach makes no assumptions about listener classes.
- Other build systems like `buck` and `bazel` can generate their equivalent of `R2` very easily by implementing the logic on their own or by just reusing `FinalRClassBuilder` (written entirely in java) from the gradle plugin module.

## Cons
- When performing rename refactors of `R` values in the Android Studio/IntelliJ, they are kept in sync with the layout files etc. automatically by the IDE, but `R2` is only generated after the gradle task that generates `R`. To keep `R` and `R2` in sync, the rename would need to be done twice once for `R` and immediately in `R2`. Without doing so, compilation would fail. This problem can be fixed easily by writing a custom intellij plugin that links the values in `R2` and `R`, similar to how the values in `R` are linked to the layout xml files etc. This can be a future enhancement.


hzsweers:
image

hzsweers:
screenshot 2016-06-10 22 00 26

hzsweers:
Wish the processor could have extracted the fully qualified names by itself so the plugin wasn't necessary for that, but can't have it all :). Awesome work

peacepassion:
This solution is similar as [Butterfork](https://github.com/oguzbabaoglu/butterfork).   

Users have to replace R by R2/B.  

I found a better solution [ButterCookie](https://github.com/JackCho/ButterCookie). Using ButterCookie, you don't need to change any existing code with Butterknife in Library project.

kageiit:
> I found a better solution ButterCookie. Using ButterCookie, you don't need to change any existing code with Butterknife in Library project.

I have looked at ButterCookie, but here are some issues I noticed with it
- Relies very heavily on what annotations are supported by Butterknife in its plugin class and the format of the generated code. This will break with butterknife extensions and new binding/listener annotations added to butterknife itself.
- Manipulates the `variant.javaCompiler` dsl which will not work with jack.
- Does reflection on the internal private APIs of the android gradle plugin and is very easily prone to breakage with android plugin updates.
- Relies on the internal implementation details of the `apt` plugin

It would need to keep up with every possible combination of changes in butterknife/the android gradle plugin/apt plugin and would become very hard to maintain.

kageiit:
> This solution is similar as Butterfork.
> 
> Users have to replace R by R2/B.

This is slightly different than Butterfork because
- It generates an `R2` class with integers instead of strings. This makes the IDE refactor friendlier than Butterfork, as there is no need to do a full recompile when renaming R values.
- Butterfork lacks the type safety ensured by using support annotations like `@IdRes`, `@StringRes` etc (which are part of butterknife) to make sure the integer you are passing to the annotation is of the correct type. The R2 class comes with the correct annotations on all the generated fields, so this issue does not exist in this approach.
- It also gives safety in the form of custom lint checks to make sure you do not use R2 in places not intended.
- ButterFork would need to keep in sync with butterknife always, which can get hard to maintain

oguzbabaoglu:
@kageiit Fully agreed on the points about Butterfork, I will deprecate it as soon as this gets merged. Great work! 👍

GrahamRogers:
Would it be possible to do something like this to generate `R2` using `ButterKnifeProcessor` instead of a Gradle plugin?
```java
AnnotationMirror rClassAnnotationMirror = getRClassAnnotationMirror();
Element rClass =
    (TypeMirror)
        AnnotationMirrors.getAnnotationValue(rClassAnnotationMirror, "value").getValue();
List<RInnerClass> rInnerClasses = new ArrayList<>();
for (Element element : rClass.asElement().getEnclosedElements()) {
  Name rInnerClassName = element.getSimpleName();
  List<? extends Element> rInnerClassFields = element.getEnclosedElements();
  rInnerClasses.add(new RInnerClass(rInnerClassName, rInnerClassFields));
}
// Use rInnerClasses to generate R2.
```
(See #591 for how getRClassAnnotationMirror could be implemented.)

I'm fairly new to Java processors, so I may be missing something, but if this is possible it would be preferable to the Gradle plugin.

GrahamRogers:
Also, instead of annotating a random type with `@RClass`, it might make more sense to make it target packages and use `package-info.java`.

e.g.
```java
@RClass(R.class)
package my.android.library;
```
would generate `my.android.library.R2`.

kageiit:
@grahamrogers I spent a lot of time trying to generate R2 using during the annotation processor rounds. The problem lies since R2 does not exist, The processor would need to defer processing on all the butterknife annotated elements to the next round. But then, the type mirror of R2.id.value etc would have changed on the next round, so the int value cannot be extracted easily by querying the value parameters on the butterknife annotations.

This also makes the logic inside the processor needlessly complex as it would have to handle the cases when the RClass annotation exists and not. Would be hard to maintain as butterknife evolves as well.

I settled on this approach as a nice balance between ease of use, a clean api and maintainability.

The location of where to out the RClass does not matter, but the import to R does not exist yet if it's on top of a package element. One would then have to use the fully qualified name. I would prefer to have it look terse instead. The decision of where to put the RClass annotation can be left to the consumer.

GrahamRogers:
For the Bazel case, since there is one Android library per package annotating the package would make sense for that case since the annotated package would be the same as the package used for `R` and `R2`. It's only a minor detail, but adding `PACKAGE` as a `Target` for `RClass` would at least give people the option and wouldn't change the implementation.

You raise good points about using code generation for `R2`, although I would question whether they outweigh having to maintain a plugin for each supported build system. I don't think there's a right answer, so plugins are fine for me (as long as a Bazel one is made :stuck_out_tongue:).

kageiit:
Not a big fan about adding a package target for RClass. The package for R2 is decided by the plugin of the build system so you could just put that detail in bazel to allow it. The same goes for buck.

The build systems are responsible for generating R. They should be able to generate R2 as well. I don't think a plugin will be provided out of the box for buck/bazel in butterknife. I'm planning to add first party support for R2 in buck after this merges. You may need to propose the same in bazel for your use case.

butterknife-sample-lib/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+    mavenCentral()
+  }
+  dependencies {
+    classpath "${GROUP}:butterknife-gradle-plugin:${VERSION_NAME}"
+  }
+}
+
+apply plugin: 'com.android.library'
+apply plugin: 'com.neenbedankt.android-apt'
+apply plugin: 'butterknife'
```
JakeWharton: It's unfortunate that you need both a plugin and the `@RClass` annotation. Couldn't this plugin modify the annotation processor options or register a task to generate a file with the right information?

kageiit: @RClass and plugin are two separate things. RClass is used as a hitn to the processor to find the symbol mapping. It works on apps without using the plugin.

The plugin is just used to create a R with all final values.

JakeWharton: Got it. You might be interested in [this issue][1] I filed a while back for exposing things to processors automatically.

 [1]: http://b.android.com/187527

oguzbabaoglu: Libraries depend on the plugin so no issue there. For apps that don't depend on it; "if you want nicer generated code, add the `@RClass` annotation" and "if you want nicer generated code, add the plugin" sounds like equivalent effort for the consumer.

Moving the responsibility to the plugin does mean it needs to be replicated for each build system (Buck, Bazel etc.). The `variant.javaCompile.options` method might not be as straightforward, but "generate a file with information" should be simple enough to do for each?

Native support by android-tools would definitely be awesome. Parsing the manifest and deriving information always feels a bit hacky.

kageiit:
I think the effort of having to use RClass in conjunction with the plugin is minimal. In buck, we just need to add an option for generating R2 and the rest will work the same. Modifying the annotation options as part of buck itself just to support usage of butterknife is not ideal.

The same goes for bazel. This way, there is no need to wait on android-tools for this support, since that will be purely be scoped to only gradle users. Once that support is available, we can make the usage of R2 optional as well for gradle users by enhancing the plugin.

kageiit:
We should also not rely on variant.javacompile dsl as that is not compatible with Jack. The android apt plugin currently does not work when Jack is enabled due to the same reason.

kageiit:
Also, if we go down the approach of "generating a file with information", anytime that file format changes, all build systems would need to fix their logic based on the version of butterknife they depend on which is not desirable. This decouples the logic and is single responsibility for the annotation and the plugin.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -95,21 +104,30 @@
...
+  private static final List<String> SUPPORTED_TYPES = Arrays.asList(
+      "array", "attr", "bool", "color", "dimen", "drawable", "id", "integer", "string"
+  );
+
   private Elements elementUtils;
   private Types typeUtils;
   private Filer filer;
+  private Trees trees;
+
+  private final Map<Integer, Id> symbols = new HashMap<>();
```
JakeWharton: nit: `LinkedHashMap`

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1107,4 +1132,77 @@ private static boolean isFieldRequired(Element element) {
...
+            VarScanner scanner =
+                new VarScanner(ids, packageName + ".R." + classTree.getSimpleName().toString());
+            ((JCTree) classTree).accept(scanner);
+          }
+        }
+      }
+    }
+  }
+
+  private static class VarScanner extends TreeScanner {
```
JakeWharton: Any idea how all this stuff fares in Jack?

butterknife-compiler/src/test/java/butterknife/RClassTest.java
```diff
@@ -0,0 +1,234 @@
...
+        + "public class Test$$ViewBinder<T extends Test> implements ViewBinder<T> {\n"
+        + "  @Override\n"
+        + "  public Unbinder bind(Finder finder, T target, Object source) {\n"
+        + "    Resources res = finder.getContext(source).getResources();\n"
+        + "    bindToTarget(target, res);\n"
+        + "    return Unbinder.EMPTY;\n"
+        + "  }\n"
+        + "  @SuppressWarnings(\"ResourceType\")\n"
+        + "  protected static void bindToTarget(Test target, Resources res) {\n"
+        + "    target.one = res.getInteger(test.R.integer.res);\n"
```
JakeWharton: Looks like we can clean this up so that it's not always going to use a FQCN. I can do that in a follow-up.

butterknife-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,27 @@
+apply plugin: 'groovy'
```
JakeWharton: Going to convert this to Kotlin post-merge. Groovy is toxic.

butterknife/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.library'
+apply plugin: 'com.kageiit.lintrules'
```
JakeWharton: The implementation of this plugin is a little scary!

JakeWharton: (The modifying the user's home directory part which isn't triggered here)

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
kageiit: Tested all this with jack enabled already :) It works as expected.

butterknife-compiler/src/test/java/butterknife/RClassTest.java
kageiit: Ya for sure. i didnt get time to clean that part up.

butterknife-gradle-plugin/build.gradle
kageiit: sure :) The plugin itself is rather simple so it should not be too hard

butterknife/build.gradle
kageiit: Ya that part is for apps, which we are not using here. I am planning to remove custom lint for app projects from the plugin and cut a new release soon.

JakeWharton:
First IDE import is problematic:
```
Error:Could not find com.jakewharton:butterknife-gradle-plugin:8.1.1-SNAPSHOT.
Searched in the following locations:
    file:/Applications/Android Studio 2.2.app/Contents/gradle/m2repository/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/maven-metadata.xml
    file:/Applications/Android Studio 2.2.app/Contents/gradle/m2repository/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.pom
    file:/Applications/Android Studio 2.2.app/Contents/gradle/m2repository/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.jar
    https://repo1.maven.org/maven2/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/maven-metadata.xml
    https://repo1.maven.org/maven2/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.pom
    https://repo1.maven.org/maven2/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.jar
    https://plugins.gradle.org/m2/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/maven-metadata.xml
    https://plugins.gradle.org/m2/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.pom
    https://plugins.gradle.org/m2/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.jar
    file:/Users/jw/.m2/repository/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/maven-metadata.xml
    file:/Users/jw/.m2/repository/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.pom
    file:/Users/jw/.m2/repository/com/jakewharton/butterknife-gradle-plugin/8.1.1-SNAPSHOT/butterknife-gradle-plugin-8.1.1-SNAPSHOT.jar
Required by:
    com.jakewharton:butterknife-sample-lib:8.1.1-SNAPSHOT
```
The one thing we do in other projects is keep the samples on the release version and only update them after a new release. I'll probably do that here too once the version with this is released.

JakeWharton:
merged as 20aad6c46eb2d4ffd06c19f0781e71df49c0b173

kageiit:
Buck support here: https://github.com/facebook/buck/pull/789

PR #628, opened 2016-06-20
Fix generated code

kageiit:
- Will add tests soon to cover all annotations


JakeWharton:
Wow there were a lot of these! Definitely need test cases to cover this stuff eventually to make sure we don't break this. I might also make `toString()` throw an exception just to guard against this...

Issue #616, opened 2016-06-15
NullPointerException  on buildToolsVersion 24.0.0 rc4


smallobserver:
i use ButterKnife 7.0.1  ,when I change buildToolsVersion to 24.0.0 rc4   it can't use. but i use some grammar in java8  so I can't change the buildToolsVersion

zubietaroberto:
Butterknife does not work with the Jack compiler (Required for compiling with build tools 24). Currently looking for a workaround.

JakeWharton:
Jack is not required for using build tools 24.

jaredsburrows:
@smallobserver Try using this with build tools `24.0.0` now.

JakeWharton:
> Butterknife does not work with the Jack compiler

This is not true. Replace `apt` with `annotationProcessor`.

zubietaroberto:
@JakeWharton Finally figured it out. It DOES work. But not in with the "stable" build tools. You need the gradle 2.2.0 alpha3 build tools. I couldn't get it working before because I try to keep to stable branches for most projects.

To get this working:

**./build.gradle**
```
dependencies {
        classpath 'com.android.tools.build:gradle:2.2.0-alpha3'
    }
```

**app/build.gradle**
```
apply plugin: 'com.android.application'

android {
    compileSdkVersion 24
    buildToolsVersion "24.0.0"

    ...
}

dependencies {
    compile 'com.jakewharton:butterknife:8.1.0'
    annotationProcessor 'com.jakewharton:butterknife-compiler:8.1.0'
}
```

**gradle/wrapper/gradle-wrapper.properties**
```
distributionUrl=https\://services.gradle.org/distributions/gradle-2.14-all.zip
```

PR #631, opened 2016-06-20
Updated documentation for Jack Support, Build Tools 24 and Java 8

zubietaroberto:
The installation information on the README does not explain how to use Butterknife with Jack Compiler and Build Tools 24. So I added a new section in the README that explains it. This configuration is working with Java 8 as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,56 @@ Make sure the line `apply plugin ...` is placed somewhere at the top of the file
 Snapshots of the development version are available in [Sonatype's `snapshots` repository][snap].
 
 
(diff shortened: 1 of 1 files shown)


zubietaroberto:
References #616

JakeWharton:
There are a lot of instructions here for things unrelated to configuring Butter Knife. I'll adapt your changes and merge manually to `master` since I don't think it's necessary to retain all this information.

JakeWharton:
I added 533e212e91c3d2d0d02d376ba1dfb7383550d9bb (which will show up after I make the release here shortly) which covers this.

PR #646, opened 2016-06-25
Fix unescaped File.separator on Windows environment (butterknife-gradle-plugin).

xirt4m:
On _Windows environment_ the following statement throws an exception:
```javascript
packageForR.replaceAll('\\.', File.separator)
```

``` 
Caused by: java.lang.IllegalArgumentException: character to be escaped is missing
	at java_lang_String$replaceAll$8.call(Unknown Source)
	at butterknife.plugin.ButterKnifePlugin$_apply_closure1$_closure2$_closure3$_closure4.doCall(ButterKnifePlugin.groovy:30)
	at org.gradle.api.internal.AbstractTask$ClosureTaskAction.execute(AbstractTask.java:554)
	at org.gradle.api.internal.AbstractTask$ClosureTaskAction.execute(AbstractTask.java:535)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeAction(ExecuteActionsTaskExecuter.java:80)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:61)
	... 70 more
```

After the proper escape of the `\\` file separator, the _butterknife-gradle-plugin_ can run without any errors.

### DIFF
--- a/butterknife-gradle-plugin/src/main/groovy/butterknife/plugin/ButterKnifePlugin.groovy
+++ b/butterknife-gradle-plugin/src/main/groovy/butterknife/plugin/ButterKnifePlugin.groovy
@@ -5,6 +5,7 @@ import com.android.build.gradle.LibraryPlugin
 import com.android.build.gradle.TestPlugin
 import com.android.build.gradle.api.BaseVariant
 import com.android.build.gradle.api.BaseVariantOutput
+import groovy.json.StringEscapeUtils
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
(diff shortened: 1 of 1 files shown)


butterknife-gradle-plugin/src/main/groovy/butterknife/plugin/ButterKnifePlugin.groovy
```diff
@@ -5,6 +5,7 @@ import com.android.build.gradle.LibraryPlugin
 import com.android.build.gradle.TestPlugin
 import com.android.build.gradle.api.BaseVariant
 import com.android.build.gradle.api.BaseVariantOutput
+import groovy.json.StringEscapeUtils
```
JakeWharton: I would prefer not to use Groovy classes since I'll be abandoning the language for something more sane very soon. Can we do something a bit more manual instead?

xirt4m: These two can also work:
- [Option one](https://github.com/xirt4m/butterknife/commit/810b3bc04a0ad53863c210902c3728680908bcdb) with the [Matcher.quoteReplacement(String s)](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#quoteReplacement-java.lang.String-).
- [Option two](https://github.com/xirt4m/butterknife/commit/50fe0bb8405ac7bd0cc7a6772a25e4996779bac7) manual escape, if necessary.

Do you like one of them?

JakeWharton: I decided I don't care.

JakeWharton:
Thanks!

PR #650, opened 2016-06-27
Use compilesWithoutWarnings in tests

ChristianBecker:
I have no idea if you want to merge this, but IMO this makes sense as we want to generate code that compiles without warnings.

If you are fine with these changes, I'll squash the different commits into a single one.

### DIFF
--- a/butterknife-compiler/src/test/java/butterknife/BindViewTest.java
+++ b/butterknife-compiler/src/test/java/butterknife/BindViewTest.java
@@ -2,13 +2,13 @@ package butterknife;
 
 import butterknife.compiler.ButterKnifeProcessor;
 import com.google.testing.compile.JavaFileObjects;
-import java.util.Arrays;
 import javax.tools.JavaFileObject;
 import org.junit.Test;
 
 import static com.google.common.truth.Truth.assertAbout;
 import static com.google.testing.compile.JavaSourceSubjectFactory.javaSource;
 import static com.google.testing.compile.JavaSourcesSubjectFactory.javaSources;
+import static java.util.Arrays.asList;
 
 public class BindViewTest {
   @Test public void bindingView() {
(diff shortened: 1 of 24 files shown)


butterknife-compiler/src/test/java/butterknife/BindViewTest.java
```diff
@@ -445,6 +445,8 @@
 
     assertAbout(javaSource()).that(source)
         .processedWith(new ButterKnifeProcessor())
+        // found raw type: test.Test
+        //   missing type arguments for generic class test.Test<T>
```
JakeWharton: File a bug? Or this will never get fixed

butterknife-compiler/src/test/java/butterknife/OnClickTest.java
```diff
@@ -659,8 +660,9 @@
     );
 
     assertAbout(javaSource()).that(source)
+        .withCompilerOptions("-Xlint:-processing")
```
JakeWharton: Why is this only added on some of the tests?

ChristianBecker: This compiler option is only necessary for sources that contain unprocessed annotations (like `ColorInt`, `Nullable`, `Optional`, …). Without [this line](http://stackoverflow.com/a/26533843/1891728) this test would fail with
```
java.lang.AssertionError: Expected 0 warnings, but found the following 1 warnings:
No processor claimed any of these annotations: android.support.annotation.ColorInt
```

If you want I can add this line to all tests to be consistent throughout the codebase.

butterknife-compiler/src/test/java/butterknife/BindViewTest.java
ChristianBecker: #651

JakeWharton:
Thanks

Issue #431, opened 2016-01-21
Resource List/Array Support


georgelin422:
Is it possible to have `FieldCollectionResourceBinding` to support binging multiple resources as `List` or `Array`
so `@BindColor({R.color.red, R.color.green, R.color.blue}) int[] mColors;` can be possible in Butterknife?

This is really useful since it can eliminate some switch cases in the code.

JakeWharton:
Only if you promise me you'll stop using Hungarian notation!

georgelin422:
lol, ok I promise to be more modern
I just took a look of the source code and found it is more work than I thought to have this feature added. Anyway I will try to achieve it :)

MONIKA-KUMAR-JETHANI:
Hi @JakeWharton , @giolin , Is this fixed?

PR #654, opened 2016-06-29
Added @BindColors, @BindDimens and @BindStrings annotations

untalfranfernandez:
As referred here: https://github.com/JakeWharton/butterknife/issues/431 I've added a couple of new annotations to be able to fetch a list of resources at once.

I've decided to keep several commits in order to see how to add new annotations easily.

### DIFF
--- a/butterknife-compiler/src/test/java/butterknife/BindStringsTest.java
+++ b/butterknife-compiler/src/test/java/butterknife/BindStringsTest.java
@@ -1,0 +1,136 @@
+package butterknife;
+
+import com.google.testing.compile.JavaFileObjects;
+
+import org.junit.Test;
+
+import javax.tools.JavaFileObject;
+
+import butterknife.compiler.ButterKnifeProcessor;
+
+import static com.google.common.truth.Truth.assertAbout;
+import static com.google.testing.compile.JavaSourceSubjectFactory.javaSource;
+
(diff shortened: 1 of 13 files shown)


butterknife-compiler/src/test/java/butterknife/BindStringsTest.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  @Test public void ListTypeHasToBeString() {
+    JavaFileObject source = JavaFileObjects.forSourceString("test.Test", ""
+            + "package test;\n"
+            + "import android.app.Activity;\n"
+            + "import java.util.List;\n"
+            + "import butterknife.BindStrings;\n"
+            + "public class Test extends Activity {\n"
+            + "  @BindStrings({1, 2}) List<Integer> strings;\n"
```
kageiit: can we just use android.R or create a fake R.java for use in tests so we never break the codegen from referring to the codeblocks instead of inlined ints.

fjfdeztoro: Sure, I can change it, the only reason I had to do it that way was to keep consistency with other tests we already had. Said that, should I change it?

w1570187062:
Hi,can also add `@BindDrawables `annotations?

PR #655, opened 2016-06-29
DRY: Reuse Finder.castView(…)

ChristianBecker:


JakeWharton:
This moves the location of the actual cast into the call site and thus outside of the try/catch block which provides more information should the cast.

ChristianBecker:
Sorry, I don't get it. The only difference between the implementation of `castView` and those two versions of `find*ViewAsType` is the usage of `cls.cast(view)` instead of `(T) view`. If the cast fails, the exact same `IllegalStateException` will be thrown – or am I missing something?

JakeWharton:
If you look at the bytecode you'll see the cast doesn't actually happen
when you just do `(T) whatever`. It happens at the call-site of that method
instead.

ChristianBecker:
Ok, so maybe we should change the implementation of `castView` to also rely on `cls.cast(view)` so that the remaining usages of `castView` can benefit from the `IllegalStateException`, too?

Issue #669, opened 2016-07-11
8.2.0 Lint issue - SuspiciousImport


jaredsburrows:
In `8.0.0` we also had a lint issue. Here is `8.2.0` lint issue:

```
    <>/Activity_ViewBinding.java:4: Error: Don't include android.R here; use a fully qualified name for each usage instead [SuspiciousImport]
    import android.R;
    ~~~~~~~~~~~~~~~~~

       Explanation for issues of type "SuspiciousImport":
       Importing android.R is usually not intentional; it sometimes happens when
       you use an IDE and ask it to automatically add imports at a time when your
       project's R class it not present.

       Once the import is there you might get a lot of "confusing" error messages
       because of course the fields available on android.R are not the ones you'd
       expect from just looking at your own R class.

    1 errors, 0 warnings
    Wrote HTML report to file:///<>/build/outputs/lint-results-debug.html
    Wrote XML report to file:///<>/build/outputs/lint-results-debug.xml
    :lintDebug FAILED

    FAILURE: Build failed with an exception.

    * What went wrong:
    Execution failed for task ':lintDebug'.
            > Lint found errors in the project; aborting build.

            Fix the issues identified by lint, or add the following to your build script to proceed with errors:
    ...
    android {
        lintOptions {
            abortOnError false
        }
    }
    ...

    * Try:
    Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

            BUILD FAILED
```

JakeWharton:
This is already fixed by #665. I didn't realize it was a lint error though. That's really dumb. I guess I'll make a point release for it.

jaredsburrows:
@JakeWharton Thanks for the quick response! I'll stick with `8.1.0` for now. Do you mind if I make a pull request for test app with lint turned on in order to capture these?

JakeWharton:
The sample is decoupled from the code on master because it now depends on
the Gradle plugin, so it can't help catch problems anymore.

jaredsburrows:
SuspiciousImport is actually a lint warning. I just force them as errors to ensure there is no issues on releases.

Here is the lint configuration I use:
```
    lintOptions {
        textReport true
        textOutput "stdout"
        checkAllWarnings true
        warningsAsErrors true
    }
```

Here is your test app: https://github.com/JakeWharton/butterknife/blob/master/sample/app/build.gradle#L22.

[...]

PR #682, opened 2016-07-19
Butterknife integration test

jaredsburrows:
@JakeWharton Based on our conversation here: https://github.com/JakeWharton/butterknife/issues/669#issuecomment-233482890.

I combined both `:samples:app` and `:sample:library` source code as `:butterknife-integration-test` and turned on `lint`.

To test my integration module, you can simple swap out:
```
  compile project(':butterknife')
  apt project(':butterknife-compiler')
```
with:
```
  compile 'com.jakewharton:butterknife:8.2.0'
  apt 'com.jakewharton:butterknife-compiler:8.2.0'
```
and you will see the "SuspiciousImport" warning.


butterknife-integration-test/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+    textOutput 'stdout'
+  }
+
+  buildTypes {
+    debug {
+      minifyEnabled true
+    }
+  }
+
+  lintOptions {
```
jaredsburrows: Turn on lint

jaredsburrows:
@JakeWharton Anything else you wanted?

butterknife-integration-test/src/androidTest/java/com/example/butterknife/SimpleActivityTest.java
```diff
@@ -0,0 +1,15 @@
+package com.example.butterknife;
+
+import android.test.ActivityInstrumentationTestCase2;
+import com.example.butterknife.library.SimpleActivity;
+
+public final class SimpleActivityTest extends ActivityInstrumentationTestCase2<SimpleActivity> {
```
JakeWharton: These can all be removed from the sample so that they only exist here.

butterknife-integration-test/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+
+  defaultConfig {
+    applicationId 'com.example.butterknife'
+    minSdkVersion rootProject.ext.minSdkVersion
+    targetSdkVersion rootProject.ext.targetSdkVersion
+    versionCode 1
+    versionName '1.0.0'
+  }
+
+  lintOptions {
```
serj-lotutovici: This one is unnecessary, you have the same block below.

butterknife-integration-test/src/test/java/com/example/butterknife/unbinder/UnbinderTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.robolectric.RobolectricGradleTestRunner;
+import org.robolectric.RuntimeEnvironment;
+
+import static com.google.common.truth.Truth.assertThat;
+
+@RunWith(RobolectricGradleTestRunner.class)
+public final class UnbinderTest {
+
+  @Test
+  public void verifyContentViewBinding() {
```
serj-lotutovici: formating: should be `@Test public void verifyContentViewBinding()`

jaredsburrows:
@serj-lotutovici @JakeWharton I fixed both issues.

JakeWharton:
It doesn't look like you deleted these files from the sample, otherwise they would appear as moved instead of new in the diff.

jaredsburrows:
@JakeWharton You commented on this file: `.../java/com/example/butterknife/SimpleActivityTest.java
`. What other files do you want removed? We need to keep at least these two files: `SimpleAdapter` and `SimpleActivity`. I fixed the lint errors in both. So if anyone introduces any, the build will fail.

JakeWharton:
Every test file

jaredsburrows:
@JakeWharton I now have deleted both the `androidTest` and `test` folders.

JakeWharton:
I wanted them deleted from the sample, not from the integration test.

jaredsburrows:
@JakeWharton I now have moved both the `androidTest` and `test` folders from the `sample` into the `butterknife-integration-test`. Is there anything else?

PR #702, opened 2016-08-09
Remove boiler plates from ButterKnife

idosu:
I added a functionality that allows the user to remove overriding onCreate and calling ButterKnife.bind from Activity and Fragment, You can look to the changes I have made to README.md I think it will explain this better,

Thank you,
Ido

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,19 @@ class ExampleActivity extends Activity {
     // TODO Use fields...
   }
 }
+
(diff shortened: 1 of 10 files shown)


JakeWharton:
This library has made a decision to not take part in view inflating at all. It was discussed long ago in #8. Furthermore, this solution uses runtime reflection which seems counter to the spirit of code generation that's currently in use for everything.

PR #709, opened 2016-08-19
let @BindView(s) play nicely with elsewhere generated Views

dmccartney:
Right now you can't `@BindView` a generated view type.

```java
@BindView(R.id.thing) GeneratedView thing;
``` 

The current compiler doesn't generate anything for an annotation on a field of a type that doesn't exist yet:
```java
// Process each @BindView element.
for (Element element : env.getElementsAnnotatedWith(BindView.class)) {
  if (!SuperficialValidation.validateElement(element)) continue;
  ...
```

This PR makes the `@BindView(s)` processing more permissive: if the view type is unresolved when the butterknife processor runs, then it will still generate the binder and note that it is relying on the type to be elsewhere generated as a `View` or interface. (But if the view type _is_ resolved, then it still performs all the same `View`/interface type checking as before.)


JakeWharton:
Hmm I thought that generation of these types would trigger another processing round since their used on an annotated element, but I guess that only makes sense if the generated code _contains_ annotations elements that this processor would handle.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -257,7 +257,8 @@
 
     // Process each @BindView element.
     for (Element element : env.getElementsAnnotatedWith(BindView.class)) {
-      if (!SuperficialValidation.validateElement(element)) continue;
+      // we don't SuperficialValidation.validateElement(element)
```
JakeWharton: nit: capitalize W and end with a period. [Comment should be sentences][1]. Same for the next comment.

 [1]: nedbatchelder.com/blog/201401/comments_should_be_sentences.html

JakeWharton:
Cool. I'm a bit worried about removing the superficial validator since it guards against other broken code problems, but we'll see what happens. I'd rather it work for the case you clearly have.

dmccartney:
Thank you :)

PR #728, opened 2016-09-03
Update to robolectric 3.1.2

jaredsburrows:
@JakeWharton Update to Robolectric 3.1.2

### DIFF
--- a/butterknife-integration-test/build.gradle
+++ b/butterknife-integration-test/build.gradle
@@ -42,4 +42,5 @@ dependencies {
   testCompile deps.junit
   testCompile deps.truth
   testCompile deps.robolectric
+  testCompile deps.openglApi
 }
--- a/build.gradle
+++ b/build.gradle
@@ -72,7 +72,8 @@ ext.deps = [
     // Test dependencies
     junit: 'junit:junit:4.12',
     truth: 'com.google.truth:truth:0.28',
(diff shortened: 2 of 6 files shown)


butterknife/src/test/java/butterknife/internal/UtilsTest.java
```diff
@@ -14,7 +15,7 @@
 import static org.junit.Assert.fail;
 
 @RunWith(RobolectricTestRunner.class)
-@Config(manifest = Config.NONE)
+@Config(constants = BuildConfig.class)
```
JakeWharton: Why do we need this and the resource file?

butterknife-integration-test/build.gradle
```diff
@@ -42,4 +42,5 @@ dependencies {
   testCompile deps.junit
   testCompile deps.truth
   testCompile deps.robolectric
+  testCompile deps.openglApi
```
JakeWharton: Why?

jaredsburrows: A bug when using API 23: https://github.com/robolectric/robolectric/issues/1932.

butterknife/src/test/java/butterknife/internal/UtilsTest.java
jaredsburrows: @JakeWharton Reverted.

PR #731, opened 2016-09-03
Update support libraries 24.2.0

jaredsburrows:
@JakeWharton 
- Updated android gradle plugin 2.1.3(didn't see this before)
- Updated google android library to 4.1.1.4
- Updated support libraries to 24.2.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,12 +51,13 @@ ext {
 }
 
 def androidToolsVersion = '25.1.2'
+def supportLibraryVersion = '24.2.0'
 
 ext.deps = [
     // Android
     android: 'com.google.android:android:4.1.1.4',
-    supportv4: 'com.android.support:support-v4:24.1.0',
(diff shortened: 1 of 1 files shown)


jaredsburrows:
@JakeWharton Starting with support library `24.2.0`, the new min sdk is 9. We can make sure we are not using an methods that are API 8 and override this or bump butterknife to min sdk 9.

ChristianBecker:
Do we really need the complete `support-v4` or can we depend on a support-split directly?

jaredsburrows:
@ChristianBecker It is only used if the developer has it in their dependencies. See the `provided` configuration here: https://github.com/JakeWharton/butterknife/blob/master/butterknife/build.gradle#L23.

JakeWharton:
We only need support-compat. Planning to fix it later.

jaredsburrows:
@ChristianBecker @JakeWharton I can change it to `support-compat`.

PR #732, opened 2016-09-03
Update butterknife in sample to 8.4.0

jaredsburrows:
@JakeWharton Updated the samples to use the latest butterknife

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,6 +52,7 @@ ext {
 
 def androidToolsVersion = '25.1.2'
 def supportLibraryVersion = '24.2.0'
+def butterknifeLatestReleaseVersion = '8.4.0'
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -51,6 +51,7 @@ ext {
 }
 
 def androidToolsVersion = '25.1.2'
+def butterknifeVersion = '8.4.0'
```
JakeWharton: This should not be in the root build file. It conflates the meaning of what's in `gradle.properties`

jaredsburrows: I can rename it to "latestReleaseVersion"?

JakeWharton: That's fine

jaredsburrows:
@JakeWharton Fixed.

PR #788, opened 2016-10-26
Scope symbol IDs per element package name

### LINKED ISSUE #779: ButterKnife generates ViewBinding class with wrong resource ID
I am using a 3rd-party library in our project, and compilation works fine, but the app crashes when it starts up. I looked at the logcat and generate source code, it seems the generated source code has the wrong resource ID.

For example, the source file is

```
class HomeActivity ... {
    @BindView(R2.id.mode_transition_layout) AppModeTransitionLayout modeTransitionLayout;

    @Override
    protected void onCreate(@Nullable Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        setContentView(R.layout.activity_main);
        ButterKnife.bind(this);
```

The generated `HomeActivity_ViewBinding` class has

```
  @UiThread
  public HomeActivity_ViewBinding(T target, View source) {
    this.target = target;

    target.modeTransitionLayout = Utils.findRequiredViewAsType(source, R.id.infant_description_text, "field 'modeTransitionLayout'", AppModeTransitionLayout.class);
  }
```

`R.id.infant_description_text` is a different resource ID, and I'd assume here should use `R.id.mode_transition_layout`. 

At the runtime I get an exception says
`java.lang.IllegalStateException: Required view 'infant_description_text' with ID 2131822441 for field 'modeTransitionLayout' was not found. If this view is optional add '@Nullable' (fields) or '@Optional' (methods) annotation.`

We are using ButterKnife 8.4.0.

Any help is highly appreciated!

felipecsl:
the integer ID itself is not enough to correctly identify a symbol. We need to scope that somehow in order to prevent clashes. This can be observed in a multiple modules project where butterknife is used in modules that depend on other modules that also use Butterknife.
Since IDs can repeat across modules, we need to scope them according to the package name of the current element being evaluated.
This PR fixes #779


butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1231,17 +1246,20 @@ private void scanForRClasses(RoundEnvironment env) {
       for (Element element : env.getElementsAnnotatedWith(annotation)) {
         JCTree tree = (JCTree) trees.getTree(element, getMirror(element, annotation));
         if (tree != null) { // tree can be null if the references are compiled types and not source
+          String respectivePackageName =
+              elementUtils.getPackageOf(element).getQualifiedName().toString();
+          scanner.setCurrentPackageName(respectivePackageName);
```
felipecsl: This is pretty janky, but couldn't figure out a simpler way to do this. Suggestions welcome

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1284,28 +1302,38 @@ private void parseCompiledR(TypeElement rClass) {
   }
 
   private static class RClassScanner extends TreeScanner {
-    private final Set<String> rClasses = new LinkedHashSet<>();
+    // Maps the currently evaulated packageName to R Classes
+    private final Map<String, String> rClasses = new LinkedHashMap<>();
```
felipecsl: This is also a bit confusing

felipecsl:
@kageiit wonder if you could take a look too, since you made this change originally

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1215,11 +1230,11 @@ private static AnnotationMirror getMirror(Element element,
     return null;
   }
 
-  private Id getId(int id) {
-    if (symbols.get(id) == null) {
-      symbols.put(id, new Id(id));
+  private Id getId(QualifiedId qualifiedId) {
```
kageiit: QualifiedId already has the id as part of it right? why do you still need the symbol map

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1231,17 +1246,20 @@ private void scanForRClasses(RoundEnvironment env) {
...
 
-    for (String rClass : scanner.getRClasses()) {
-      parseRClass(rClass);
+    for (Map.Entry<String, String> rClass : scanner.getRClasses().entrySet()) {
+      parseRClass(rClass.getKey(), rClass.getValue());
     }
   }
 
-  private void parseRClass(String rClass) {
+  private void parseRClass(String respectivePackageName, String rClass) {
```
kageiit: rename to rPackageName?

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
felipecsl: Correct. I kept the Map just for the fast lookup

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
felipecsl: sure!

felipecsl:
@JakeWharton friendly ping!

JakeWharton:
Monday at earliest. Traveling.

felipecsl:
No worries, I just realized that this is not quite right yet. More updates coming

kunong:
@felipecsl anymore updates coming? I'm a guy that waiting this fix. :)

felipecsl:
@kunong yep, planning to finish today or tomorrow

javajoker:
@felipecsl , any update on this issue? Thanks, Ning

felipecsl:
no updates yet. We stopped seeing this issue so it's not burning anymore.

mrhether:
@felipecsl what exactly still needs to be completed here? Once we know, then hopefully someone can work on it.

mrhether:
@felipecsl what still needs to be done here?

felipecsl:
@mrhether If I'm not mistaken, this change breaks usage of ButterKnife when you're not using it as a gradle plugin. If you always use it as a plugin, then it should work fine.

mrhether:
@felipecsl Is this hosted anywhere?

felipecsl:
Nope

felipecsl:
Alright thanks to @BenSchwab 's https://github.com/airbnb/butterknife/commit/b0aa4ad417ab0867a8bbd19c5d9461c835996e6f we this is finally functional and ready for review.
We've been using it internally for a ~week and it seems to have fixed the bug.
@JakeWharton PTAL at your earliest convenience 😄

build.gradle
```diff
@@ -37,7 +37,7 @@ subprojects { project ->
       }
     }
     dependencies {
-      classpath 'com.android.tools.build:gradle:2.2.2'
+      classpath 'com.android.tools.build:gradle:2.3.0-rc1'
```
JakeWharton: Is this required? Non-final releases are timeboxed and I don't want to commit code that uses pre-release versions because that makes the commit unbuildable after a month or two.

JakeWharton:
Looks like there's a checkstyle failure.

build.gradle
felipecsl: not required. it was just Android Studio being annoying. Reverting now..

felipecsl:
Green and ready to go 🎉

PR #790, opened 2016-11-02
Fix listener method type checking

### LINKED ISSUE #782: @OnTextChanged crashes on string
```
@OnTextChanged(R.id.foo) void onTextChanged(String bar) {
}
```

This crashes when the generated code attempts to cast the CharSequence to a String, but the CharSequence isn't actually a String.

Either Butterknife should complain about String as a type here, or it should call toString() on the CharSequence.

NikoYuwono:
Fix #782 

- Add checking to listener method parameter to allow using subtype of original method's type only when the parameter is subtype of view.

- Add testcase in OnTextChangedTest.java

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,7 @@ subprojects { project ->
   version = VERSION_NAME
 
   repositories {
+    mavenCentral()
     maven {
       url "https://plugins.gradle.org/m2/"
     }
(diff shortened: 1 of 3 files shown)


NikoYuwono:
Looks like `com.android.tools.lint:lint-tests:25.1.2` can't be found in `https://plugins.gradle.org/m2/` if add `https://dl.bintray.com/android/android-tools` to repositories, it builds with no problem.

JakeWharton:
Add `mavenCentral()`. It shouldn't have been removed in the first place.

JakeWharton:
Thanks!

PR #797, opened 2016-11-08
Move more to immutable architecture.

JakeWharton:


JakeWharton:
This was in my work tree. No clue when I wrote it... tests pass so whevs.

PR #812, opened 2016-11-22
[Lint] Migrate to new PSI API

felipecsl:
I was able to confirm that the ButterKnife custom lint was the reason for this issue with Retrolambda to resurface: https://github.com/evant/gradle-retrolambda/issues/96
Apparently, if you have any custom lint rules using the old lombok API, the linter will run it twice, both with the old and with the new API, making the lint slower and also causing it to spit enourmous amounts of stacktraces due to the retrolambda incompatibility, as described in the issue linked above.
Unfortunately, the custom lombok ast build stopped working for reasons I can't understand, but fixing the butterknife linter should be enough to have the linter stop spitting stacktraces (as long as you don't have any other custom lint rules using the lombok api).
This PR migrates the custom R2 lint rule to the new PSI API
Sorry if the description got confusing, it's 2:35am 😴

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ ext {
   targetCompatibilityVersion = JavaVersion.VERSION_1_7
 }
 
-def androidToolsVersion = '25.1.2'
+def androidToolsVersion = '25.2.0'
 def supportLibraryVersion = '24.2.1'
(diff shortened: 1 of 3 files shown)


felipecsl:
@kageiit should probably take a look too, since you wrote the original rule

PR #843, opened 2017-01-09
Add SuppressLint("ClickableViewAccessibility") to OnTouchListener's

### LINKED ISSUE #772: Apply @SuppressLint attribute to generated handler methods when necessary
I have my `lint.xml` set up so that `ClickableViewAccessibility` raises an `error`. 

The `@OnTouch` binding does not have a way to suppress that warning when I am making sure to call `performClick` when necessary in my annotated method.

I recommend either picking up `@SupressLint` annotations on bound methods and applying them to the generated listener, or adding an optional string parameter to the `@On[Event]` annotations themselves for this situation.

ChristianBecker:
Closes #772

### DIFF
--- a/butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
+++ b/butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
@@ -1,5 +1,6 @@
 package butterknife.compiler;
 
+import butterknife.OnTouch;
 import butterknife.internal.ListenerClass;
 import butterknife.internal.ListenerMethod;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
```diff
@@ -542,6 +551,16 @@ private boolean hasMethodBindings() {
     return false;
   }
 
+  private boolean hasOnTouchMethodBindings() {
+    for (ViewBinding bindings : viewBindings) {
+      if (bindings.getMethodBindings()
+          .containsKey(OnTouch.class.getAnnotation(ListenerClass.class))) {
```
ChristianBecker: I'm open to suggestions if you are not happy with the explicit dependency on the OnTouch annotation

butterknife/src/test/java/butterknife/OnTouchTest.java
```diff
@@ -33,6 +34,7 @@
         + "  private Test target;\n"
         + "  private View view1;\n"
         + "  @UiThread\n"
+        + "  @SuppressLint(\"ClickableViewAccessibility\")\n"
```
ChristianBecker: If you want I can move this annotation directly to the `onTouch(..)` implementation to reduce the scope.

JakeWharton: Yes I would prefer this, if it's not too difficult.

ChristianBecker: Unfortunately, it's not that easy without extending `ListenerMethod` or without including another hard-coded check for `OnTouch` in `BindingSet#addMethodBindings(..)`. What do you think?

JakeWharton: Ok. Not a big deal.

JakeWharton:
Thanks!

PR #872, opened 2017-02-10
Generate R2 entries for all resource types

### LINKED ISSUE #871: Allow layout resources in R2
Currently the gradle plugin skips layout types when generating R2 (https://github.com/JakeWharton/butterknife/blob/master/butterknife-gradle-plugin/src/main/java/butterknife/plugin/FinalRClassBuilder.java#L28)

Butterknife does not use these, but I have a feature in [Epoxy](https://github.com/airbnb/epoxy) where a layout resource is a value in an annotation. I would love to support library projects, without having to duplicate the plugin just to add support for layout.

I know it's strange to request this from Butterknife, since it isn't related to the library, but the gradle plugin Butterknife provides is potentially useful to many other libraries. It could be its own generic library.

Would it be possible to have the plugin take parameters in our gradle files to specify which resource types it should allow?

@kageiit R2 actually does include layout files when building with Buck. I'm wondering if you did something different with buck to allow all resources?

kageiit:
- Fixes #871

### DIFF
--- a/butterknife-gradle-plugin/src/main/java/butterknife/plugin/FinalRClassBuilder.java
+++ b/butterknife-gradle-plugin/src/main/java/butterknife/plugin/FinalRClassBuilder.java
@@ -12,8 +12,6 @@ import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.TypeSpec;
 import java.io.File;
-import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


simplysoft:
We tried this, and it fails for "mipmap" resource types, as it generates a @MipmapRes annotation that does not exist

JakeWharton:
Merged #969.

PR #877, opened 2017-02-16
Update License year

JEKos:


JakeWharton:
We don't do this internally for our projects at Square which are overseen by our legal team so I'm inclined to not touch it for any of my personal projects. There's also a scary clause in copyright law that says something to the effect of changing copyright incorrectly can invalidate it. Thanks, though.

PR #891, opened 2017-03-02
Fix README.md example using R2 instead of R

RenanLukas:


JakeWharton:
That's only for libraries.

PR #898, opened 2017-03-13
Correct typecast checking utility.

kbsriram:
Fix utility that checks for parameter types so it correctly
checks and reports type errors.


JakeWharton:
Actually I might revert this. These are checked at compile-time, no?

PR #922, opened 2017-04-05
use getClassLoader().loadClass() replace for Class.forName , because …

jiawulu:
for our atlas project , business has bundle ClassLoader , if the ButterKnife is in maindex, the Class.forName can't load the class in the bundle . so if you can use class.getClassLoader() instead , thanks


JakeWharton:
Thanks!

Issue #870, opened 2017-02-09
Bind animation resources


diogobernardino:
I couldn't find a way to bind animation resources. Is it provided? Or is it something considered to be implemented?

JakeWharton:
Send a PR? Also are you talking about the old view animations or the new
property animators? Or both?

diogobernardino:
Sounds like a no. Not talking about one in particular: view, or property, or both.

PR #934, opened 2017-04-25
Added support for binding animation resources

aurae:
Related to the request in #870, this PR adds a new `@BindAnim` annotation, with which `Animation` resources can be bound to fields.

Example Usage:

```java
public class Test {
    @BindAnim(R.anim.fade_in)
    Animation fadeIn;
}
```

Generated code uses `AnimationUtils#loadAnimation` to retrieve the resource.


sp00ne:
This is a great addition to butterknife imo. With the bonus of loadAnimation being static to begin with :)

rocboronat:
+1!

PR #950, opened 2017-05-08
Merge buildscript blocks.

JakeWharton:


jaredsburrows:
`:sample:library` needs a repository as it brings in `classpath "com.jakewharton:butterknife-gradle-plugin:${versions.release}"`.

PR #951, opened 2017-05-10
bind bean into views

taoisym:
I want implement feture like this
class BeanFromServer{
         public  String name;
         public  String name(){
                 return name;
         };
}
@BindBeanClass(BeanFromServer.class)
class Fragment{
@BindBean(R.id.name,"name")
     TextView name;    
}
//then then innder class UnBinder will add a  ***apply*** function
//and add a field ***Binder binder***
Binder binder;
public void apply(BeanFromServer bean){
     binder.apply(name,bean.name);
     //depend on BindBean annotationed on ,and some other type can be generate,like this
    binder.apply(name,bean.name());
    binder.apply(name,func(bean.name));
    binder.apply(name,func(bean.name()));
}

the binder have default implement  and can be rewrite and custom
interface Binder{
     <Target,Data>void apply(Target target,Data data);
    .....
}


JakeWharton:
I think you want to use databinding from the Android team. I don't want to go down the rabbit hole of binding data back into UI.

Thanks for taking the time to make the PR though.

https://developer.android.com/topic/libraries/data-binding/index.html

PR #969, opened 2017-06-05
Generate R2 entries for anim, layout, menu, plurals, style, and styleable

### LINKED ISSUE #871: Allow layout resources in R2

ngsilverman:
This is a change we've been using at Airbnb for Epoxy as well as for an internal library that uses style and styleable resources.

Similar to #872 but explicitly adds resource types as well as the relevant tests.

Fixes #871, a request to add layout resource types to R2 as they're used by [Epoxy](https://github.com/airbnb/epoxy).

cc @kageiit @elihart


ngsilverman:
@JakeWharton any interest in adding this to the Butterknife plugin or should other libraries provide their own?

gogile:
I would be obliged if you could kindly merge it.

PR #976, opened 2017-06-15
Color tint support for @BindDrawable

sp00ne:
Hi there Jake!

I'm a great fan of your work :) Something that I noticed was the need for color resource support in the tinting of drawables, the pull request adds `@ColorRes` annotation for the tint binding and makes the necessary changes to the code for it to work. I expanded the sample activity to include an `ImageView` that is tinted in this manner.

I hope this meets your standards.

Best regards,
Mohsen Farzone


JakeWharton:
I'm not sure that I want to conflate this attribute with two meanings. Is there a reason you don't want to define the color you're tinting with on the theme?

sp00ne:
Yeah, in most cases where I work (especially larger projects) the colors are not associated with the theme or there is a situational basis for the coloring

ivanick87:
hskxu

sp00ne:
Further, do you have the checkstyle so I can make sure it passes the check? :)

PR #1026, opened 2017-07-28
Update to AGP 3.0.0a8

jrodbx:
- Bump to Gradle 4.0-M1
- Bump to AGP 3.0.0-alpha8
- Update dependency configuration syntax
- Tell Robolectric to include Android resources in tests
- Update plugin usage of variant api

### DIFF
--- a/butterknife-compiler/build.gradle
+++ b/butterknife-compiler/build.gradle
@@ -1,4 +1,4 @@
-apply plugin: 'java'
+apply plugin: 'java-library'
 apply plugin: 'checkstyle'
 
 sourceCompatibility = JavaVersion.VERSION_1_7
@@ -13,14 +13,14 @@ for (File file : sdkHandler.sdkLoader.repositories) {
 }
 
 dependencies {
-  compile project(':butterknife-annotations')
-  compile deps.auto.common
-  compile deps.javapoet
+  implementation project(':butterknife-annotations')
+  implementation deps.auto.common
+  api deps.javapoet
   compileOnly deps.auto.service
   compileOnly files(org.gradle.internal.jvm.Jvm.current().getToolsJar())
 
-  testCompile deps.junit
-  testCompile deps.truth
+  testImplementation deps.junit
+  testImplementation deps.truth
 }
 
 checkstyle {
--- a/settings.gradle
+++ b/settings.gradle
@@ -5,7 +5,7 @@ include ':butterknife-gradle-plugin'
 include ':butterknife-lint'
 include ':butterknife-integration-test'
 
-include ':sample:app'
-include ':sample:library'
+//include ':sample:app'
+//include ':sample:library'
 
 rootProject.name = 'butterknife-parent'
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
       'buildTools': '26.0.0',
 
       'supportLibrary': '26.0.0',
-      'androidPlugin': '2.3.3',
+      'androidPlugin': '3.0.0-alpha8',
       'androidTools': '25.3.0',
       'kotlin': '1.1.3-2',
 
--- a/butterknife-annotations/build.gradle
+++ b/butterknife-annotations/build.gradle
@@ -1,4 +1,4 @@
-apply plugin: 'java'
+apply plugin: 'java-library'
 apply plugin: 'checkstyle'
 
 sourceCompatibility = JavaVersion.VERSION_1_7
@@ -11,7 +11,7 @@ checkstyle {
 
 dependencies {
   compileOnly deps.android.runtime
-  compile deps.support.annotations
+  api deps.support.annotations
 }
 
 apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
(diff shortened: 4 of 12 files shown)


jrodbx:
```
* What went wrong:
A problem occurred configuring project ':sample:library'.
> Failed to notify project evaluation listener.
   > com.android.build.gradle.api.BaseVariant.getOutputs()Ljava/util/List;
```

This error happens locally as well.  It seems that :sample:library has a buildscript dependency on the previously published version of butterknife-gradle-plugin, which fails with 3.0.0-alpha8 due to a variant api change, so we have a 🐔 & 🥚.  Essentially, this build won't pass unless a new version of the plugin is published.

To test locally, I did the following:
1) In gradle/gradle-mvn-push.gradle, update urls to local repo
```
  repository(url: mavenLocal().url) {
  snapshotRepository(url: mavenLocal().url) {
```
2) Build and install updated plugin to local repository
```
  ./gradlew :butterknife-gradle-plugin:uploadArchives
```
3) In sample/library/build.gradle, add mavenLocal() to buildscript.repositories
```
  buildscript {
    repositories {
      ...
      mavenLocal()
    }
  }
```
4) Also, in sample/library/build.gradle, update plugin version to local SNAPSHOT
```
  dependencies {
    classpath "com.jakewharton:butterknife-gradle-plugin:SNAPSHOT"
  }
```
5) Finally, build and test
```
  ./gradlew :sample:app:build
```

JakeWharton:
Just disable the sample from settings.gradle

jrodbx:
works for me.  updated.

butterknife-compiler/build.gradle
```diff
@@ -13,14 +13,14 @@ for (File file : sdkHandler.sdkLoader.repositories) {
 }
 
 dependencies {
-  compile project(':butterknife-annotations')
-  compile deps.auto.common
-  compile deps.javapoet
+  implementation project(':butterknife-annotations')
+  implementation deps.auto.common
+  api deps.javapoet
```
JakeWharton: This is implementation too, no? We don't expose these types in the public API.

jrodbx: maybe?  i changed it to `api` in response to this error:
```
> Task :butterknife:compileDebugUnitTestJavaWithJavac
error: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider butterknife.compiler.ButterKnifeProcessor could not be instantiated: java.lang.NoClassDefFoundError: com/squareup/javapoet/ClassName
```

JakeWharton: This is because of how we're running the compiler in the tests. Just going to leave it for simplicity.

jrodbx: 👍

denizmveli:
@jrodbx this error is still occurring for me using butterknife 8.8.0 in one of our library projects. It's using `alpha9`.

```
* What went wrong:
A problem occurred configuring project ':mobile'.
> Could not resolve all dependencies for configuration ':mobile:kapt'.
   > A problem occurred configuring project ':player'.
      > Failed to notify project evaluation listener.
         > com.android.build.gradle.api.BaseVariant.getOutputs()Ljava/util/List
```

I've cleaned all caches and re-run. Is there something obvious I'm missing?

JakeWharton:
Use 9.0.0-SNAPSHOT

Frank1234:
Could you update the snapshot to beta1? "Gradle sync failed: Could not find com.android.tools.build:gradle:3.0.0-alpha9."

JakeWharton:
That's an issue on your side unless Google removed alpha9 from their repo
(which seems unlikely).

jrodbx:
Just tested and 3.0.0-alpha9 is still available. @Frank1234 did you include google() in your buildscript.repositories?  That's a necessary step now if you happen to be migrating from 2.3.3 to 3.0.0*.

Frank1234:
thanks @jrodbx, that's it

settings.gradle
```diff
@@ -5,7 +5,7 @@ include ':butterknife-gradle-plugin'
 include ':butterknife-lint'
 include ':butterknife-integration-test'
 
-include ':sample:app'
-include ':sample:library'
+//include ':sample:app'
```
jaredsburrows: @jrodbx @JakeWharton Did you mean to leave this commented out?

jrodbx: yep.  see https://github.com/JakeWharton/butterknife/pull/1026#issuecomment-318657996

jaredsburrows:
Thanks. I was looking into solutions earlier to use the plugin module
directly (buildSrc module, jar directly, snapshot, etc). The buildSrc
solution is the only one that solves it entirely, I might make a PR later.

AndSync:
remove  apply plugin: 'com.jakewharton.butterknife' may be worked

butterknife-compiler/build.gradle
TWiStErRob: @JakeWharton `implementation deps.javapoet; testRuntimeOnly deps.javapoet` so the consumers are not affected, and it reflects the intent?

JakeWharton: Sure. Send a PR

PR #1043, opened 2017-08-04
Remove old stub from test code.

JakeWharton:


JakeWharton:
Works locally. Whevs.

Issue #808, opened 2016-11-18
Generated ViewBinding doesn't extend from library module ViewBinding


sosite:
There's problem with extend views (Activities, Fragments, etc.) from library module and overriding inflated layout. Unfortunately your great, etc. library doesn't catch this kind of situations so generated `ViewBinding` for one module doesn't extends from another one.

In example situation `ActivityB_ViewBinding` should extends from `ActivityA_ViewBinding`

```java
public class ActivityB_ViewBinding<T extends ActivityB> implements Unbinder {
```
```java
public class ActivityA_ViewBinding<T extends ActivityA> implements Unbinder {
```

should be
```java
public class ActivityB_ViewBinding<T extends ActivityB> extends ActivityA_ViewBinding<T> {
```
```java
public class ActivityA_ViewBinding<T extends ActivityA> implements Unbinder {
```

app module:
---
```java
public class ActivityB extends ActivityA {
    @BindView(R.id.text_b) TextView textB;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        textB.setText("test");
    }

    @LayoutRes
    @Override
    protected int getLayoutRes() {
        return R.layout.activity_b;
    }
}
```
gradle includes:
```gradle
annotationProcessor 'com.jakewharton:butterknife-compiler:8.4.0'
compile project(':lib')
```

lib module:
---
```java
public abstract class ActivityA extends Activity {
    @BindView(R2.id.text_a) TextView textA;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(getLayoutRes());
        ButterKnife.bind(this);
        // textA is null when ActivityA is extended in other module!
        textA.setText("test");
    }

    @LayoutRes
    protected int getLayoutRes() {
        return R.layout.activity_a;
    }
}
```
gradle includes:
```gradle
compile 'com.jakewharton:butterknife:8.4.0'
annotationProcessor 'com.jakewharton:butterknife-compiler:8.4.0'
```

---
`activity_b.xml` layout of course have 2 `TextView` with `R.id.text_a` and `R.id.text_b`.  Unfortunatelly ButterKnife doesn't have anything against when we forget to include layout with `R.id.text_a` - generator validator doesn't work at compile time here.

logs:
```
Looking up binding for com.test.ActivityB
HIT: Loaded binding class and constructor.
```

[...]

PR #1045, opened 2017-08-04
support inheritance in modules

stephanenicolas:
Solves issue #808:

This PR:
* fixes issue #808. The BK AP can use a super class that was already compiled in a different module. It checks for the view binder of such classes, and looks at some extra properties of the binder to ensure the subclasses are generated correctly (whether they need a context to bind view or not).
* add a simple inheritance unit test. It uses a base class that is precompiled to simulate that the super class of a class under test could already have been compiled in a different module.
* changes the sample (here, I am wondering how you guys do to change the sample before releasing..)

We would appreciate your feedback on this PR about both the idea of the solution and its implementation, tests, etc.


tongngocdat:
shopquannamdangcap

pumba210:
Hi

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -367,6 +369,12 @@
         bindingMap.put(type, builder.build());
       } else {
         BindingSet parentBinding = bindingMap.get(parentType);
+
+        // parent binding is null, let's try to find a previouly generated binding
```
opsenes: Typo in comment -> "previouly"

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -380,6 +388,33 @@
     return bindingMap;
   }
 
+  @NonNull private BindingSet createStubBindingSet(TypeElement parentType) {
+    BindingSet parentBinding;
+    BindingSet.Builder parentBuilder = BindingSet.newBuilder(parentType);
+    if (hasViewBindings(parentType)) {
+      //add a fake field to the parent class so that it will indicate it has a view bindings.
+      //this is required for the subclass to generate a proper view binder
```
oldergod: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

PR #1075, opened 2017-09-12
Make ButterknifeProcessor use significantly lesser memory during compilation

kageiit:
Currently, Butterknife processor scans and parses all available `R` class files and their resources. When working with modules which have a large number of dependencies, the many merged `R` files from different packages and same resource identifiers lead to significant amount of memory required to maintain mappings of ids.

- This change makes Butterknife only maintain mapping of ids that are actually referenced in the various annotations.
- Some element instances have been converted to not be serialized to strings until necessary.

Doing so consumes significantly lesser memory and leads to faster compilation as well. Some snapshots of heap usage during compilation before and after

Before
=====
old_heap_uage
old_retained_heap

After
====
new_heap_usage
new_retained_heap

The retained heap after the change is due to another [bug](https://github.com/typetools/checker-framework/issues/1482) not related to Butterknife.

As for build time, one of our largest modules (~2.5k java files) that uses butterknife modules went from 130s to 90s to build (about *30 %* savings)


JakeWharton:
Merged manually with some indentation fixes. Thanks.

PR #1109, opened 2017-10-13
Accept all SDK licenses without having to maintain a hash

sridoodla:


JakeWharton:
I tried a bunch of combinations last night to get this to work. It's a mess. Basically you have to use `sdkmanager` to get an updated version of the tools which has the `--licenses` flag which means you need to echo the license anyway 🤦‍♂️ .

sridoodla:
Ah, I didn't think about that. I just did this while building a docker image for our android CI builder and it worked great.

But because I had complete control over the version of the tools ( and I used the latest ), I never thought about that.

JakeWharton:
Yep. Hopefully https://issuetracker.google.com/issues/64934388 will help
for users of Travis CI. I don't really want to get in the business of
having to mess with an entire Docker image for stuff like this.

PR #1122, opened 2017-10-26
Removed support.annotations dependency from Butterknife

Macarse:
While learning the difference between api/implementation I noticed that:

 -  `butterknife-annotations` exposes `deps.support.annotations`
 - `butterknife` depends on `butterknife-annotations`
 - `butterknife` has `implementation deps.support.annotations`

So, `butterknife` doesn't need `deps.support.annotations`.

Does it make sense?

### DIFF
--- a/butterknife/build.gradle
+++ b/butterknife/build.gradle
@@ -27,7 +27,6 @@ android {
 
 dependencies {
   api project(':butterknife-annotations')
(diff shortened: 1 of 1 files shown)


JakeWharton:
Probably just historical baggage. Makes no difference in the end so :+1:

Issue #1124, opened 2017-10-27
NoSuchMethodError: com.android.build.gradle.tasks.ProcessAndroidResources.getPackageForR()Ljava/lang/String;


scottkennedy:
```
Caused by: java.lang.NoSuchMethodError: com.android.build.gradle.tasks.ProcessAndroidResources.getPackageForR()Ljava/lang/String;
        at butterknife.plugin.ButterKnifePlugin$configureR2Generation$1$1.execute(ButterKnifePlugin.kt:59)
        at butterknife.plugin.ButterKnifePlugin$configureR2Generation$1$1.execute(ButterKnifePlugin.kt:18)
```

It seems that AGP 3.1 has removed a method that the plugin uses.

JakeWharton:
Can use `getOriginalApplicationId` but it'll also break eventually.

amrfarid140:
I am facing the same error. How can I workaround it?

JakeWharton:
Use 3.0.0

amrfarid140:
Sorry but 3.0.0 of what, Android Studio?

JakeWharton:
Android Gradle plugin

dazza5000:
Thank you

kassim:
I'm getting this stack trace on building after updating to `com.android.tools.build:gradle:3.1.0-alpha05`


```
java.lang.NoSuchMethodError: com.android.build.gradle.tasks.ProcessAndroidResources.getOriginalApplicationId()Ljava/lang/String;
        at butterknife.plugin.ButterKnifePlugin$configureR2Generation$1$1.execute(ButterKnifePlugin.kt:59)
        at butterknife.plugin.ButterKnifePlugin$configureR2Generation$1$1.execute(ButterKnifePlugin.kt:18)
        at org.gradle.api.internal.DefaultDomainObjectCollection.all(DefaultDomainObjectCollection.java:136)
        at butterknife.plugin.ButterKnifePlugin$configureR2Generation$1.execute(ButterKnifePlugin.kt:52)
        at butterknife.plugin.ButterKnifePlugin$configureR2Generation$1.execute(ButterKnifePlugin.kt:18)
        at org.gradle.internal.ImmutableActionSet$CompositeSet.execute(ImmutableActionSet.java:196)
        at org.gradle.internal.MutableActionSet.execute(MutableActionSet.java:35)
[...]

PR #1128, opened 2017-10-30
For android gradle plugin 3.1.0-alpha01, update the butterknife gradl…

CaMnter:
**Android gradle plugin 3.1.0-alpha01 Version**:
In `ProcessAndroidResources`, there is no `packageForR` field. Will throw an exception:

```shell
Error:A problem occurred configuring project ':app'.
> Could not determine the dependencies of task ':app:processDebugResources'.
   > Could not resolve all dependencies for configuration ':app:debugCompileClasspath'.
      > A problem occurred configuring project ':widget'.
         > Failed to notify project evaluation listener.
            > com.android.build.gradle.tasks.ProcessAndroidResources.getPackageForR()Ljava/lang/String;
```

This will cause gradle compilation to fail.

---

In the **Android gradle plugin 3.0 version** `ProcessAndroidResources` source, found a section on the `packageForR` assignment code:

```java
processResources.packageForR =
                        TaskInputHelper.memoize(
                                () -> {
                                    String splitName = config.getSplitFromManifest();
                                    if (splitName == null) {
                                        return config.getOriginalApplicationId();
                                    } else {
                                        return config.getOriginalApplicationId() + "." + splitName;
                                    }
                                });
```

With reference to this code, by way of reflection, access to the `ProcessAndroidResources` field. And then set the value of the **R package**, making the normal operation.

This is when the android gradle plugin is upgraded, causing some fields to lose the problem. So, I think here can be used to reflect the way. Just what I think.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
       'compileSdk': 26,
 
       'supportLibrary': '27.0.0',
-      'androidPlugin': '3.0.0',
+      'androidPlugin': '3.1.0-alpha01',
       'androidTools': '26.0.0',
       'kotlin': '1.1.51',
 
--- a/butterknife-gradle-plugin/src/main/java/butterknife/plugin/ButterKnifePlugin.kt
+++ b/butterknife-gradle-plugin/src/main/java/butterknife/plugin/ButterKnifePlugin.kt
@@ -1,12 +1,9 @@
 package butterknife.plugin
 
-import com.android.build.gradle.AppExtension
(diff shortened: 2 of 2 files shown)


JakeWharton:
This is not a suitable fix as it adds more dependencies onto the internals of the plugin. See https://github.com/JakeWharton/butterknife/issues/1124#issuecomment-340124969 for what we should fall back to with reflection.

CaMnter:
Oh, I know, thank you.

PR #1162, opened 2017-12-15
Remove android:components block

Jawnnypoo:
As mentioned, can be removed since 3.0.0 of the plugin is out. I figured this is a good thing to do here since others might use this `travis.yml` as a template

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,5 @@
 language: android
 
-android:
-  components:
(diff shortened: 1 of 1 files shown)


JakeWharton:
These actually used to be important because Travis isn't good at updating their SDK. We actually had to install newer version of tools so that we could install newer versions of platform-tools in order for things like lint, sdkmanager, and the like to work. I guess Travis has updated the default SDK since then...

Jawnnypoo:
Ahh good to know, thanks!

PR #1168, opened 2017-12-28
ButterKnife.findById methods are deprecated, don't encourage using them

TWiStErRob:
https://github.com/JakeWharton/butterknife/blob/master/CHANGELOG.md#version-880-2017-08-04
> Deprecate the findById methods. Compile against API 26 and use the normal findViewById for the same functionality.


JakeWharton:
Thanks. I need to pull this in on master first as that's the source of truth for the website.

TWiStErRob:
Oh, should've looked at the commits, I'll rebase.

PR #1176, opened 2018-01-08
Use `BaseVariant.getApplicationId()` to make ButterknifePlugin compatible with AGP 3.1.0-alpha07

raviagarwal7:
Use `BaseVariant.getApplicationId()` instead of `ProcessAndroidResources.getOriginalApplicationId()` which was moved to an internal class `LinkApplicationAndroidResourcesTask` in `AGP 3.1.0-alpha07`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
       'compileSdk': 27,
 
       'supportLibrary': '27.0.2',
-      'androidPlugin': '3.0.1',
+      'androidPlugin': '3.1.0-alpha07',
       'androidTools': '26.0.1',
       'kotlin': '1.2.10',
 
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
       'compileSdk': 27,
 
       'supportLibrary': '27.0.2',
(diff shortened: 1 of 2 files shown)


hzsweers:
Does `BaseVariant.getApplicationId()` work on AGP 3.0?

raviagarwal7:
yes it does, `getApplicationId()` has always been there on `BaseVariant`. I am not sure why `ProcessAndroidResources.getOriginalApplicationId()` was used in the first place.

If `getOriginalApplicationId` is necessary then will have to ask the AGP team to add this back to ProcessAndroidResources public api.

build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
       'compileSdk': 27,
 
       'supportLibrary': '27.0.2',
-      'androidPlugin': '3.0.1',
+      'androidPlugin': '3.1.0-alpha07',
```
JakeWharton: This should always be a stable version.

raviagarwal7: okay, will keep the version to be 3.0.1

JakeWharton:
Thanks!

PR #1181, opened 2018-01-17
Stringify resource IDs using Integer.toHexString to avoid issues with hyphens for resource IDs larger than 0x7FFFFFFF.

donaldchai:
Resource IDs such as 0x80000000 were previously stringified using as signed decimals, leading to invalid Java identifiers such as "view-2147483648".


JakeWharton:
Looks like the integration test failed but CI doesn't show why. I'll have to try this locally.

donaldchai:
Ah, sorry, I think I just ran the integration tests.  Looks like there are some field names in UnbinderTest.java which need to be updated as well.

JakeWharton:
Thanks!

PR #1194, opened 2018-02-03
ButterKnife.findById methods are deprecated, don't encourage using them

TWiStErRob:
https://github.com/JakeWharton/butterknife/blob/master/CHANGELOG.md#version-880-2017-08-04
> Deprecate the findById methods. Compile against API 26 and use the normal findViewById for the same functionality.

(Exactly the same as #1168, same branch, same base, GitHub's "change base" doesn't seem to work, sorry about the mess there.)

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -202,14 +202,6 @@ void onNothingSelected() {
   // TODO ...
 }</pre>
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
Thanks!

PR #1207, opened 2018-02-23
Set API level in LintRegistry.

runningcode:
This avoids a warning when running lint with new versions of AGP.

Also update some dependency versions.

### DIFF
--- a/butterknife-lint/src/main/java/butterknife/lint/LintRegistry.java
+++ b/butterknife-lint/src/main/java/butterknife/lint/LintRegistry.java
@@ -1,6 +1,7 @@
 package butterknife.lint;
 
 import com.android.tools.lint.client.api.IssueRegistry;
+import com.android.tools.lint.detector.api.ApiKt;
 import com.android.tools.lint.detector.api.Issue;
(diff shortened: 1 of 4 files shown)


butterknife-lint/src/main/java/butterknife/lint/LintRegistry.java
```diff
@@ -13,4 +13,8 @@
   @Override public List<Issue> getIssues() {
     return ImmutableList.of(InvalidR2UsageDetector.ISSUE);
   }
+
+  @Override public int getApi() {
+    return com.android.tools.lint.detector.api.ApiKt.CURRENT_API;
```
JakeWharton: Import?

runningcode: haha, i copied it straight from the lint warning. fixed!

runningcode:
Aww, this is causing a build failure: https://issuetracker.google.com/issues/73450622

jrodbx:
Fixed by: https://github.com/JakeWharton/butterknife/pull/1208

PR #1214, opened 2018-03-01
optimization for function findViewById

stormagain:
If findViewByld is called with high frequency, there is a overhead of recursions. Let's make it faster. Update includes new implementation of findViewById with Map used for memoization for previous search result.


JakeWharton:
What problem is this solving?

stormagain:
Total time consumption，when we have too many views need find out.

JakeWharton:
How are you measuring? Systrace? What's the depth of the hierarchy and how many views are you looking up, for example?

I'm really hesitant about adding something like this without knowing more.

stormagain:
Thank you for reminding, I have maked some ABTest on my devices locally, there is no difference between the two ways. next time I'll do the test first，sorry to disturb.

PR #1250, opened 2018-04-12
Fix R scanner behavior for array values

### LINKED ISSUE #1247: ButterKnife @BindViews not working correctly.
Hey, I'm trying to create a list of views with ButterKnife Binding but the library not working as expected.
This is my code where I'm creating list of views.

`@BindViews({R.id.trackerHistoryExpandImageView, R.id.trackerHistoryExtraLayout})
 List<View> viewList;
`

Library correctly made list of views but when I debug my code I got the error in library.
This is the ButterKnife generated code looks like for the list of views.

`
target.viewList = Utils.listOf(
        Utils.findRequiredView(source, R.id.trackerHistoryExtraLayout, "field 'viewList'"), 
        Utils.findRequiredView(source, R.id.trackerHistoryExtraLayout, "field 'viewList'"));
`

You see the views Utils class is finding, the same view for both time "R.id.trackerHistoryExtraLayout";

This is my library version for butterknife.

`
// Butter knife dependencies
    implementation 'com.jakewharton:butterknife:9.0.0-SNAPSHOT'
    annotationProcessor 'com.jakewharton:butterknife-compiler:9.0.0-SNAPSHOT'
`

Any help would be highly appreciable.

kageiit:
Fixes #1247 

- TreeScanner api visits all values at once when calling `accept(Tree)` so aggregate the symbols within the scanner itself
- Added new unit tests

### DIFF
--- a/butterknife-compiler/src/main/java/butterknife/compiler/FieldDrawableBinding.java
+++ b/butterknife-compiler/src/main/java/butterknife/compiler/FieldDrawableBinding.java
@@ -25,7 +25,7 @@ final class FieldDrawableBinding implements ResourceBinding {
   }
 
   @Override public CodeBlock render(int sdk) {
-    if (tintAttributeId.value != 0) {
+    if (tintAttributeId.value != -1) {
       return CodeBlock.of("target.$L = $T.getTintedDrawable(context, $L, $L)", name, UTILS, id.code,
           tintAttributeId.code);
     }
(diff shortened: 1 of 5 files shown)


butterknife-compiler/src/main/java/butterknife/compiler/FieldDrawableBinding.java
```diff
@@ -25,7 +25,7 @@
   }
 
   @Override public CodeBlock render(int sdk) {
-    if (tintAttributeId.value != 0) {
+    if (tintAttributeId.value != -1) {
```
sunian: this `-1` is kind of a magic number. it's not clear to me from looking at this class why the value would ever be -1.
perhaps make a `NO_ID` constant in `BindDrawable`?

kageiit: Was thinking about the same. But BindDrawable may not be the best place for it

kageiit:
Seems like the `connectedCheck` task is failing (Its passing locally for me on sdk 26 emulator)

kageiit:
Instrumentation tests worked on an api 19 emulator as well :| . Maybe travis ci has some flakiness?

kageiit:
rekicking travis ci

Issue #1280, opened 2018-05-15
Android X support


xTheITx:
I've recently attempted to upgrade a project which uses Butterknife v8.8.1 to make use of the new Android X libraries by using the jetifier tool in Android Studio. 

The project fails to compile as the generated *_ViewBinding classes reference the old android.support.annotation.* annotations which no longer exist. 
There may also be additional issues I'm not aware of.

I was just wondering if there was a plan to migrate over to reference the new androidx package names in the generated code. 

I'm not familiar with the Butterknife code, but let me know if there's anything I could contribute to the project to facilitate adding support. 

Thank you for your time

scarrie:
I got my app compiling by adding my own copies of the old package "android.support.annotation"s for @CallSuper and @UiThread.

`@Retention(CLASS)
@Target({METHOD,CONSTRUCTOR,TYPE,PARAMETER})
public @interface UiThread {
}`

`@Retention(CLASS)
@Target({METHOD})
public @interface CallSuper {
}`

Its a start, but then I hit further problems caused by this new jetifier process with other libraries. Guess we just have to be patient to wait for all these third party libraries to update.

allidoiswarren:
I'll take a crack at this.

justinegeo96:
I'm sorry to say the error still persists.

naturalwarren:
Can you paste the stack trace? Are you building against a JAR cut from master?

ArunPandiyan:
It's from the import statement for the annotations which is found in the auto generated viewbinding files.

k4dima:
repositories {
        maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
    }
    implementation 'com.jakewharton:butterknife:9.0.0-SNAPSHOT'
    annotationProcessor 'com.jakewharton:butterknife-compiler:9.0.0-SNAPSHOT'

feinstein:
>repositories {
    maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
}
implementation 'com.jakewharton:butterknife:9.0.0-SNAPSHOT'
annotationProcessor 'com.jakewharton:butterknife-compiler:9.0.0-SNAPSHOT'

This isn't working for me

adityathakar:
Is it safe to use 9.0.0-SNAPSHOT version in production code?

SyedNabil:
[...]

PR #1288, opened 2018-06-07
Add AndroidX Support

allidoiswarren:
Addresses #1280 

It would be nice if we could get this functionality tested within ButterKnife. Should we move ButterKnife to AGP 3.2 and create a Jetifier enabled app as an integration test?

For the time being I was able to deploy to maven local and successfully built an AndroidX app using this change. Strangely I had to declare an explicit dependency on ButterKnife's annotations artifact otherwise they didn't appear on my classpath.


allidoiswarren:
Got access to my other account back. Going to add the PR from there.

PR #1289, opened 2018-06-07
Add AndroidX support.

naturalwarren:
Addresses #1280

It would be nice if we could get this functionality tested within ButterKnife. Not sure the best way to go about since we have to add androidx artifacts to the processor classpath.

For the time being I was able to deploy to maven local and successfully built an AndroidX app using this change. Strangely I had to declare an explicit dependency on ButterKnife's annotations artifact otherwise they didn't appear on my classpath.


naturalwarren:
Travis timed out but this builds just fine.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -202,6 +202,7 @@
   private Map<TypeElement, BindingSet> findAndParseTargets(RoundEnvironment env) {
     Map<TypeElement, BindingSet.Builder> builderMap = new LinkedHashMap<>();
     Set<TypeElement> erasedTargetNames = new LinkedHashSet<>();
+    boolean useAndroidX = hasAndroidX();
```
JakeWharton: Can we do this in `init` instead?

butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
```diff
@@ -76,6 +85,7 @@ private BindingSet(TypeName targetTypeName, ClassName bindingClassName, boolean
     this.isView = isView;
     this.isActivity = isActivity;
     this.isDialog = isDialog;
+    this.useAndroidX = useAndroidX;
```
JakeWharton: This seems like something we should inject into `brewJava` since it's not a property of the things being bound but rather an implementation detail of _how_ the bindings are generated, just like the SDK level and whether or not the variant is debugging.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1348,6 +1354,21 @@ private static AnnotationMirror getMirror(Element element,
...
 
+  /**
+   * Perform two lookups to see if the androidx annotation and core libraries are on the classpath.
+   * If both aren't present butterknife will leverage support annotations and compat libraries
+   * instead.
+   */
+  static boolean hasAndroidX() {
+    try {
+      Class.forName("androidx.annotation.NonNull");
+      Class.forName("androidx.core.content.ContextCompat");
```
JakeWharton: These are unlikely to be on the same classpath as the annotation processor. We should look them up from `elements.getTypeElement` to use the application classpath.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
naturalwarren: f365e09

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
naturalwarren: 92c115c

butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
naturalwarren: a479d06

butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
```diff
@@ -82,13 +88,16 @@ private BindingSet(TypeName targetTypeName, ClassName bindingClassName, boolean
     this.parentBinding = parentBinding;
   }
 
-  JavaFile brewJava(int sdk, boolean debuggable) {
-    return JavaFile.builder(bindingClassName.packageName(), createType(sdk, debuggable))
+  JavaFile brewJava(int sdk, boolean debuggable, boolean useAndroidX) {
+    return JavaFile.builder(bindingClassName.packageName(), createType(
```
JakeWharton: nit: can we pull the `createType` call into a local. I really don't like the wrapping here.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -666,9 +669,10 @@ private void parseResourceColor(Element element,
     int id = element.getAnnotation(BindColor.class).value();
     Id resourceId = elementToId(element, BindColor.class, id);
     BindingSet.Builder builder = getOrCreateBindingBuilder(builderMap, enclosingElement);
-    builder.addResource(new FieldResourceBinding(resourceId, name,
-        isColorStateList ? FieldResourceBinding.Type.COLOR_STATE_LIST
-            : FieldResourceBinding.Type.COLOR));
+    builder.addResource(new FieldResourceBinding(resourceId, name, isColorStateList ? (useAndroidX
+        ? FieldResourceBinding.Type.COLOR_STATE_LIST_ANDROIDX
+        : FieldResourceBinding.Type.COLOR_STATE_LIST) : (useAndroidX
+        ? FieldResourceBinding.Type.COLOR_ANDROIDX : FieldResourceBinding.Type.COLOR)));
```
JakeWharton: Double nested ternary is hurting my brain. Can we pull each branch of the outermost ternary into a local so that the result is just `isColorStateList ? colorStateListType : colorType`?

butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
naturalwarren: 34b4a43

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
naturalwarren: 24afa90

JakeWharton:
Great work. Thanks!

butterknife-compiler/src/main/java/butterknife/compiler/BindingSet.java
AamirNaseer1: how can I add support for AndroidX

PR #1291, opened 2018-06-11
Update build.gradle

SnehPandya18:
Change google() repository & jcenter() repository access priority so that the gradle build would not fail to resolve dependencies as per latest Android Studio update.

Reference: https://stackoverflow.com/questions/50584437/android-studio-3-1-2-failed-to-resolve-runtime/50758769

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -69,8 +69,8 @@ subprojects { project ->
   version = VERSION_NAME
 
   repositories {
-    google()
     mavenCentral()
+    google()
     jcenter()
   }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -69,9 +69,9 @@ subprojects { project ->
   version = VERSION_NAME
 
   repositories {
+    google()
```
JakeWharton: This should come after `mavenCentral()`

SnehPandya18: Thanks for pointing out, I updated it. 👍

JakeWharton:
Thanks

PR #1294, opened 2018-06-20
Enable task caching for R2Generator.

runningcode:
This enables gradle task caching for the R2 Generating task.

Some more info on task caching:
https://github.com/gradle/task-output-cache-demos/blob/master/docs/making-custom-tasks-cacheable.md


JakeWharton:
I don't think we can update the plugin because of this (dumb) bug: https://issuetracker.google.com/issues/73450622

runningcode:
Downgraded!

pumba210:
Hi

PR #1319, opened 2018-08-21
update gradle version

isfaaghyth:


JakeWharton:
We cannot update because aapt2 has a stupid change that prevents single-element package names. Please star https://issuetracker.google.com/issues/73450622 for updates.

isfaaghyth:
I see, thanks, Jake.

enigmatix:
Not trying to revive this, just adding some context for others who run into this problem and can better find it in search.  This is effectively the cause behind the error:

`attribute 'package' in <manifest> tag is not a valid Android package name: 'butterknife’.`

This first surfaced for us when we upgraded bazel from `0.28.1` > `3.7.1`.  Bazel removed the backwards compatability flag (allowing you to switch off aapt2) at version 2.

The solution was to move from using an aar to directly importing the classes.jar contained within.  This meant there was no xml file for aapt2 to process, and no error.

Issue #1016, opened 2017-07-19
Generate Kotlin extension functions for calling the generated ViewBindings


vanniktech:
Given a class `AboutActivity` it would be cool to have a generated extension function `AboutActivity.bind()` that just calls `AboutActivity_ViewBinding(this)`

The same could also be applied for Dialogs / Views.

What do you think?

JakeWharton:
I like it. Can we do this as part of the normal processor and just have the
kt files ignored by javac? Or do we need a second processor?

vanniktech:
Aren't `kt` files ignored by javac by default?

JakeWharton:
It's more complicated than that. If kapt will accept non-source file
outputs suffixed with .kt files then it's as easy as that, but if it
requires source file outputs and javac also tries to compile them as java
then it won't.

JakeWharton:
https://medium.com/the-fabulous/kotlin-extension-methods-generation-15b5e6499dc8

blipinsk:
@JakeWharton moving my question from reddit here:
I can create a PR with those changes, would you rather have the `ButterKnifeKtx.bind(this)` or the "original" `Activity.bindViews()` with specialized receiver type?

JakeWharton:
This will work well with the reflection-based backend, too. It means you can have direct references when the processor is used and reflection for IDE builds. No `Class.forName` anywhere.

vanniktech:
Using ViewBinding now

PR #1362, opened 2018-09-12
WIP Kotlin extension function generation

blipinsk:
*This is a followup of the #1016 issue.*

So far it is empty, to figure out details of the implementation.


blipinsk:
@JakeWharton I'm starting to work on this feature (generating extension functions).
I would like make sure we're on the same page in terms of the way you would like to have the feature implemented into the library. That's why I'm starting the PR early (without any code), mostly for the discussion purposes.

1. Base extension functions in a new module (e.g. `butterknife-ktx`) so it can be used parallely with `butterknife` or `butterknife-reflect`
2. Extension function generation included in the original `butterknife-compiler` module.

sounds good?

JakeWharton:
They don't need to be in new modules. The Kotlin dependency can be marked as optional and we can hide the enclosing class from Java callers.

blipinsk:
> They don't need to be in new modules. The Kotlin dependency can be marked as optional and we can hide the enclosing class from Java callers.

Alright, would you like the base extension functions to works differently for both "modes"? Or the same?
1. If the `base` extension function (i.e. `Activity.bind()`) is called for `butterknife`, it means that there is something wrong with the extension function generation. Throw exception?
2. If the `base` extension function is called for `butterknife-reflect` everything is correct so we're passing the call for `Butterknife.bind(this)`.

JakeWharton:
I would call `ButterKnife.bind` in both cases so that the extension is still usable for code generation from locations where you only have an `Activity` reference.

pumba210:
Hi

PR #1374, opened 2018-09-24
Speed up CI builds by using R8

JakeWharton:


pumba210:
Hi

gradle.properties
```diff
@@ -19,3 +19,5 @@ org.gradle.jvmargs=-Xmx1536M
 
 # TODO https://issuetracker.google.com/issues/73450622
 android.enableAapt2=false
+
+android.enableR8=true
```
jaredsburrows: What about `android.enableR8.fullMode=true`? See https://github.com/JakeWharton/SdkSearch/blame/master/gradle.properties#L3.

PR #1384, opened 2018-10-10
Make all artifacts use Java 8 bytecode

JakeWharton:
As promised in the RC1 release notes...


JakeWharton:
I'll put the over/under on a bug being filed about this at 18 hours. Get your votes in now!

Issue #497, opened 2016-04-24
Support Custom Annotations


YGDroid:
thanks for using butterknife.

I found butterknife anntations managed by ButterKnifeProcessor class (LISTENER list)

I want to make some custom annotation with butterknife.

let me know how I add custom annotation myself.

is it possible?

anlcnydn:
Hey,

To make custom annotations are not related with ButterKnife. It is built by using annotations. ButterKnife's annotations are also custom ones. So, yes, it is possible. ButterKnifeProcessor class extends AbstractProcessor which implements Processor interface. All of the annotations are
parsed by the Processor. You can check [this](http://programmaticallyspeaking.com/playing-with-java-annotation-processing.html) and [this](http://www.mkyong.com/java/java-custom-annotations-example/) 
for further information.

JayNewstrom:
@JakeWharton would you be willing to take a look at a PR for something along these lines if I were to work on it?

JakeWharton:
Sure!

winfoozmnayef:
HI @JakeWharton 
I need to know if i can implement this feature in ButterKnife.
That's what I'm thinking about
- Add new Annotation which is called AnnotationAdapter with `ANNOTATION_TYPE` target under butterknife-annotations module, take one attribute which is Class<?>, this annotation used for annotate custom annotation and pass implementation class to it as attribute e.g (`@AnnotationAdapter(ViewPager2Listener.class)`) the ViewPager2Listener is a class for deal with ViewPager2 and register listener to it
- Add new interface which is called ButterKnifeAdapter for example and this interface take one generic parameter which is the view you need to create custom annotation for it to use it with AnnotationAdapter
- Add another interface which is called ButterKnifeListener for example this interface to link custom listener with code generation

**Full example**

AnnotationAdapter
```java
@Target(ElementType.ANNOTATION_TYPE)
public @interface AnnotationAdapter {

    Class<? extends ButterKnifeAdapter> value();
}
```

ButterKnifeAdapter
```java
public interface ButterKnifeAdapter<T> {
}
```

ButterKnifeListener
```java
public interface ButterKnifeListener<T> {
    void invoke(T t);
}
```

OnPageChangeListener2 (My custom annotation)
```java
@AnnotationAdapter(ViewPager2Listener.class)
public @interface OnPageChangeListener2 {
    int value();
}
```

ViewPager2Listener(My custom listener)
```java
[...]

PR #1385, opened 2018-10-11
zx butterknife

zhangxing123:
clickview callback


JakeWharton:
I'm not sure what the goal is here, but #497 seems a far more scalable approach to custom listeners.

PR #1389, opened 2018-10-20
Don't assume that cached constructors are never null

### LINKED ISSUE #1388: findBindingConstructorForClass does not cache null result returned when there are no bindings
If there are no bindings in a class, but `Butterknife.bind()` is called, then it will go through `findBindingConstructorForClass` without storing the null result in the binding cache:

> D/ButterKnife: Looking up binding for com.innercirclesoftware.reddit.shared.ui.RecyclerViewSpinnerPopupAdapter$ViewHolder
> D/ButterKnife: Not found. Trying superclass com.innercirclesoftware.reddit.shared.base.BaseViewHolder
> D/ButterKnife: Not found. Trying superclass android.support.v7.widget.RecyclerView$ViewHolder
> D/ButterKnife: MISS: Reached framework class. Abandoning search.

This can happen when for example you add the `Butterknife.bind() `call to a base class which you extend from, to decrease boilerplate. Therefore, I think it would make sense to store null in the BINDINGS Map since this could be invoked many times, and is extremely expensive. For me it causes a significant lag (500ms) when attempting to bind ~20 Views with no bindings. 

Currently `Butterknife.BINDINGS` uses a `LinkedHashMap` which doesn't allow null, so it would have to be replaced to support this.

maciej-kaznowski:
Fixes #1388

Null values are already being stored in the map, so that didn't have to be updated.

### DIFF
--- a/butterknife-annotations/build.gradle
+++ b/butterknife-annotations/build.gradle
@@ -1,8 +1,8 @@
 apply plugin: 'java-library'
 apply plugin: 'checkstyle'
 
-sourceCompatibility = JavaVersion.VERSION_1_7
-targetCompatibility = JavaVersion.VERSION_1_7
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
 
 checkstyle {
   configFile rootProject.file('checkstyle.xml')
--- a/butterknife-compiler/build.gradle
+++ b/butterknife-compiler/build.gradle
@@ -1,8 +1,8 @@
 apply plugin: 'java-library'
 apply plugin: 'checkstyle'
 
-sourceCompatibility = JavaVersion.VERSION_1_7
-targetCompatibility = JavaVersion.VERSION_1_7
+sourceCompatibility = JavaVersion.VERSION_1_8
+targetCompatibility = JavaVersion.VERSION_1_8
 
 dependencies {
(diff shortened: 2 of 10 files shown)


butterknife/src/main/java/butterknife/ButterKnife.java
```diff
@@ -187,17 +187,17 @@ public static Unbinder bind(@NonNull Object target, @NonNull View source) {
 
   @Nullable @CheckResult @UiThread
   private static Constructor<? extends Unbinder> findBindingConstructorForClass(Class<?> cls) {
-    Constructor<? extends Unbinder> bindingCtor = BINDINGS.get(cls);
-    if (bindingCtor != null) {
+    if (BINDINGS.containsKey(cls)) {
       if (debug) Log.d(TAG, "HIT: Cached in binding map.");
-      return bindingCtor;
+      return BINDINGS.get(cls);
```
JakeWharton: I don't like that this requires a double hash lookup in the common case. Can we revert the logic and only check `containsKey` if `get` returns `null`?

maciej-kaznowski:
Thanks, I updated it. I am unsure however on how to test this, as it is more of an incorrect API usage case (documentation for java.util.Map.get() states that one should use Map.containsKey() to differentiate between a map storing null for the key, and a map returning a default value of null when the key isn't contained).

PR #1402, opened 2018-11-13
Support for Android Gradle Plugin 3.3.

cmwarrington:
Android Gradle Plugin includes direct bytecode generation of the R class
in libraries, which breaks previous R2 generator s it relies on the
generated source.

Add support for reading the symbol table (R.txt), as well as the R.java
source file that already existed.

Currently uses reflection on the resource processing task task to find
the value, filed  https://issuetracker.google.com/118668005 to introduce
an API in the Android Gradle Plugin.


cmwarrington:
What's the minimum AGP the butterknife gradle plugin should support?
I could remove the R.java parsing code if we're only going back to 3.0.0

JakeWharton:
I am comfortable being 3.0+. In fact, I think we're already 3.1+ because of a binary-incompatible return type change that was made between 3.0.x and 3.1.x.

JakeWharton:
Thanks I'll look soon and get a release out soon thereafter

JakeWharton:
Thanks. Will try to release in the next few days.

frogermcs:
Does anyone have a problem with R.txt file location for `com.android.feature`, in multi-module project? 

I get this message: 
```
> java.io.FileNotFoundException: .../my_app/features/{module}/build/intermediates/symbols/feature/debug/R.txt (No such file or directory)
```

It seems that Gradle Plugin 3.3 puts this file into: 
.../my_app/features/{module}/build/intermediates/symbols/debug/R.txt
(no `/feature/` part)

PR #1407, opened 2018-11-20
Explicitly check BINDINGS if it has cached a null result

maciej-kaznowski:
Following comments from https://github.com/JakeWharton/butterknife/pull/1389


JakeWharton:
Added a test and merged as e9c1533e2af2b26bf26ce13c139ae6ffdeb3a7c5

Issue #1440, opened 2019-01-27
Butterknife plugin doesn't use lazy task APIs of gradle 4.9+


stephanenicolas:
Gradle 4.9 introduced Task Configuration Avoidance:
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

Butterknife doesn't use the new APIs and trigger the creation of most of tasks for most variants.

I believe the only thing to change would be : https://github.com/JakeWharton/butterknife/blob/master/butterknife-gradle-plugin/src/main/java/butterknife/plugin/ButterKnifePlugin.kt#L64

The best would be to support older versions of gradle still: 
https://github.com/melix/jmh-gradle-plugin/blob/a034aa88805b7a06fa9c5a825d573554b2aa23e2/src/main/groovy/me/champeau/gradle/JMHPlugin.groovy#L289-L296

PR #1445, opened 2019-01-31
Support Task Configuration Avoidance

stephanenicolas:
Solves issue #1440. The change is limited.

* delay the creation of the processResourcesTask used in the generateR2 task.
* the generate R2 task itself still has to be created (because AGP doesn't allow to pass it as a task provider, I created a ticket for that).

### DIFF
--- a/butterknife-reflect/build.gradle
+++ b/butterknife-reflect/build.gradle
@@ -23,7 +23,7 @@ android {
 
   // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
   libraryVariants.all {
-    it.generateBuildConfig.enabled = false
+    it.generateBuildConfigProvider.configure { it.enabled = false }
   }
 }
 
--- a/butterknife-runtime/build.gradle
+++ b/butterknife-runtime/build.gradle
@@ -29,7 +29,7 @@ android {
 
   // TODO replace with https://issuetracker.google.com/issues/72050365 once released.
   libraryVariants.all {
-    it.generateBuildConfig.enabled = false
+    it.generateBuildConfigProvider.configure { it.enabled = false }
   }
 }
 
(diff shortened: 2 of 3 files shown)


butterknife/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="butterknife"/>
+<manifest package="butterknife.core"/>
```
stephanenicolas: This was a side effect of upgrading the AGP version, it looks like single word package names are not allowed anymore. But I don't think it has any impact.

JakeWharton: Blocked by https://issuetracker.google.com/issues/73450622

stephanenicolas: What is the impact of the change of the manifest ?

build.gradle
```diff
@@ -6,15 +6,14 @@ buildscript {
       'compileSdk': 28,
 
       'androidTools': '26.2.0',
-      'kotlin': '1.2.71',
-
+      'kotlin': '1.3.11',
```
stephanenicolas: this version of kotlin is not the latest, but it is recommended in AS 3.3.0 that goes with AGP 3.3.0

stephanenicolas:
The PR has been rebased against current master, I squashed my commits, it should now get green. The kotlin version better matches Android Studio.

stephanenicolas:
I just realized I was missing an obsolete call. It couldn't determine exactly how to test it. 
Please test it properly or let me know how to make sure the original behavior works fine.

stephanenicolas:
I've tested the second commit too and all seems to be working pretty well.

PR #1447, opened 2019-02-01
Fix usage snippet in ButterKnife Reflect README

sewar:
Using Android Studio 3.2 and Gradle 4.10.3, `properties.containsKey('android.injected.invoked.from.ide')` is always false.

### DIFF
--- a/butterknife-reflect/README.md
+++ b/butterknife-reflect/README.md
@@ -33,7 +33,7 @@ Usage
 Kotlin modules:
 ```groovy
 dependencies {
-  if (properties.containsKey('android.injected.invoked.from.ide')) {
+  if (project.hasProperty('android.injected.invoked.from.ide')) {
     implementation 'com.jakewharton:butterknife-reflect:<version>'
   } else {
     implementation 'com.jakewharton:butterknife:<version>'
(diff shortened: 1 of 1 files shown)


JakeWharton:
Hmm I'll have to test. I'm pretty sure I was using 3.3 alphas when I wrote this.

TWiStErRob:
I just spend 15 mins trying to figure out what's wrong as well. Because you're in a `dependencies` block Groovy is being clever. We need to prefix `properties` with `project.` to make sure we're reading what we're interested in, or do as @sewar suggested.

If you paste the code from README to AS, sync, and navigate to the declaration of `properties` you'll see:
```java
package org.codehaus.groovy.runtime;
public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
    public static Map getProperties(Object self);
    public static MetaProperty hasProperty(Object self, String name);
```
i.e. the `val Object.properties: Map<*, *> get()` for Kotliners.

Note that even @sewar's suggestion wouldn't work without the `project.` prefix, because it would be trying to call the above `hasProperty` method which is a Groovy reflection method.

dawidhyzy:
Any ideas on how to fix that?

TWiStErRob:
@dawidhyzy it works if you use `project.properties....` or `project.hasProperty`

EesiLife:
这是来自QQ邮箱的自动回复邮件。
 
您好，现在无法回复您的邮件。我将尽快给您回复。

PR #1448, opened 2019-02-01
Remove enforcement of required views in reflection backend

JakeWharton:
Almost all of the nullability annotations don't have runtime retention so this only causes problems. Also you'll get an NPE and it'll be obvious anyway, it just won't be eager.


JakeWharton:
@jrodbx

jrodbx:
tested and 👍

PR #1461, opened 2019-02-08
Support for generating Kotlin R2 files

ZacSweers:
This is a PR with a proposal implementation for generating Kotlin `R2` files. This would be a potential win for Kotlin consumers on the build side by not imposing a mixed source set cost.

The structure of the Kotlin versions would be simple: every class is an `object`, every field is a `const`. This is technically a bit more heavyweight than the Java equivalent since there's a singleton class being generated for each, but one would assume that all these R2 classes are being stripped in release anyway since they're only used for compile-time.

Example:
```kotlin
object R2 {
  object anim {
    @AnimRes const val fooAnim: Int = 0x7f040001
  }
}
```

This would be gated on a configurable gradle extension:

```gradle
butterKnife {
  generateKotlin = true
}
```

Opportunistically cleaned up the structure of the plugin a bit along the way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,14 @@ class ExampleActivity extends Activity {
 }
 ```
 
+By default it will generate a Java R2 class. If you want kotlin, you can configure it via the plugin
+extension:
+
+```groovy
+butterKnife {
+  generateKotlin = true
+}
+```
 
 
 License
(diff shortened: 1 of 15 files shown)


ZacSweers:
Note for the kotlin test build fixture failure, I'm not really sure what's going on as it seems like it's not pulling from the plugin classpath (and thus can't find kotlinpoet classes). Going to ask from gradle gurus for help

PR #1463, opened 2019-02-08
Update various dependencies + working nullaway

ZacSweers:
A few dependencies + usage of new lazy task configuration in AGP. Also enabled lazy task management for the gradle plugin, but will hold off on that pending how #1461 goes


ZacSweers:
Just noticed #1445

Strangely I didn't hit https://issuetracker.google.com/issues/73450622 after updating

ZacSweers:
... there it is

why doesn't this happen for me locally 🤔

JakeWharton:
Probably should split up if you want any chance of some of this landing.

Issue #1225, opened 2018-03-19
Support incremental annotation processing 


stephanenicolas:
Hi BK team, 

Gradle 4.7 provides support for incremental annotation processing. 
https://docs.gradle.org/nightly/userguide/java_plugin.html#sec:incremental_annotation_processing

It would be nice to see butter knife supporting it. The change is not that big as butter knife is an isolating AP. 

Also, to fully support incremental processing, there is a PR butter knife should merge to support inheritance a cross modules: #1045

stephanenicolas:
Closed for https://github.com/JakeWharton/butterknife/pull/1466

PR #1466, opened 2019-02-09
Support inheritance across modules and incremental annotation processing

stephanenicolas:
The PR is now ready for review: @JakeWharton @ZacSweers. 
I did my best to comment it and help your review. I remain available to address your comments and answer your questions. I would appreciate our exchanges to be respectful and open. 
Thx in advance.

This solves both: 

* support incremetal annotation processing: #1225 
* support inheritance across modules #1045 

Both issues are linked: to support incremental changes, we have to support the case where a subclass only is recompiled and the super class is not passed to the annotation processor. The same occurs for inheritance across modules.

In this case, we want to avoid doing some work that we have already done, we don't want to re-scan the annotations of the super class (it would have been possible to access it via annotation processing). To avoid this duplicate processing, we reuse the generated code of the super class in the subclass. I will comment more inside the PR on how this is currently achevied.


butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -117,13 +120,16 @@
...
+  private Elements elementUtils;
   private Filer filer;
   private @Nullable Trees trees;
 
   private int sdk = 1;
   private boolean debuggable = true;
 
   private final RScanner rScanner = new RScanner();
+  private HashMap<String, List<Element>> mapGeneratedFileToOriginatingElements
+      = new LinkedHashMap<>();
```
stephanenicolas: This is used only for testing. I will take more about tests below.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -143,10 +149,18 @@
...
-      trees = Trees.instance(processingEnv);
-    } catch (IllegalArgumentException ignored) {
+      //reflection won't be necessary after https://github.com/gradle/gradle/pull/8393
+      java.lang.reflect.Field delegateField = processingEnv.getClass().getDeclaredField("delegate");
+      delegateField.setAccessible(true);
+      trees = Trees.instance((ProcessingEnvironment) delegateField.get(processingEnv));
+    } catch (Throwable t) {
+      try {
+        trees = Trees.instance(processingEnv);
+      } catch (Throwable ignored) {
```
stephanenicolas: This whole block here is meant to access the `ProcessingEnvironment` instance passed by gradle. Gradle wraps the annotation processors internally so that it can control the filer passed to annotation processors. 

The wrapper class provided by gradle is `IncrementalProcessingEnvironment`. This class is not exposed by Gradle. The idea is to access this class and the original wrapped `ProcessingEnvironment`. We need this code because `Trees.getInstance(..)` checks that it receives the original processing environment. If you look at the code of the `Trees` class, you will see that it first checks the class name of the instance. I tried to by pass this check and call the method that `instance` is calling internally. But there are additional checks later on the road. I resorted to use reflection to access the original processing environment in the non exposed gradle wrapper class.

I have opened an issue in gradle's code base on this in: https://github.com/gradle/gradle/pull/8393
The PR has been rejected has gradle doesn't want to expose this class. As you can see in this issue, we discuss the safety of accessig `Trees` in the BK processor during incremental annotation processing. This practice is a corner case when incremental processing is used and it is generally not advised to turn annotation processors to incremental if they use this API. Though, I explained why I believe there is no issue in doing so in our case.

Also, at the end of the discussion in the gradle PR, gradle says that the processing environment is always wrapped. It was not my experience when testing: I received the vanilla processing environment. I didn't follow up with gradle on this. But for this PR, I don't think it should matter as the case was already handled by just swalling the exception. In practice, the PR has been tested at an industrial scale in 2 companies that I know of and all works well.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -201,6 +217,10 @@
     return false;
   }
 
+  public HashMap<String, List<Element>> getMapGeneratedFileToOriginatingElements() {
```
stephanenicolas: This method is also always visible for testing. Of course, we can change the visibility accordingly.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -1273,7 +1326,7 @@ private static boolean isTypeEqual(TypeMirror typeMirror, String otherType) {
         return null;
       }
       typeElement = (TypeElement) ((DeclaredType) type).asElement();
-      if (parents.contains(typeElement)) {
+      if (parents.contains(typeElement) || hasViewBinder(typeElement)) {
```
stephanenicolas: This comment is about all the changes above. Here, the idea is that we want to take into account the situation where a sub class is being processed and not its super class (because we are in an incremental build or because the super class is in a different module).

Even though the super class is not passed to the annotation processor, we can know if butterknife has generated some code for it by testing the presence of the generated class in the classpath. If such a `__Binder` exists for the super class, we create a `StubBindingSet` for it: we do re-scan its annotations here, only because we want to know which constructor of the super class to invoke later when we generate the `__Binder` of the sub class.

Of course, I am open to amend the way of doing things if there is a better solution.

butterknife-runtime/src/test/java/butterknife/BindViewTest.java
```diff
@@ -49,12 +57,21 @@
...
         .and()
         .generatesSources(bindingSource);
+
+    Map<String, List<Element>> map = butterKnifeProcessor.getMapGeneratedFileToOriginatingElements();
+    assertEquals(1, map.size());
+    Map.Entry<String, List<Element>> entry = map.entrySet().iterator().next();
+    assertEquals("test/Test_ViewBinding.java", entry.getKey());
+    List<Element> elements = entry.getValue();
+    assertEquals(1, elements.size());
+    assertEquals("test.Test", elements.get(0).asType().toString());
```
stephanenicolas: In this test, I use the method and the map I have added to the BK processor. 

There are different options here:
1) unit testing: we can test that we pass the right originating element to the filer. This is a healthy practice, idependently of gradle's incremental solution. 
2) integration testing. I've seen other projects using integration tests with gradle, with fixtures, and I am not very familiar with this technology. 

I decided to go with the first testing strategy as it limits the impact of the changes here, I also think this test and the tested behavior is just "healthier" for the compiler. But we can also prefer to test more directly that incremental annotation processing is working as expected, which would be better to ensure there is no regression at this level. It would probably bring more value for maintenance. 
If we decide to go this way though, I will need some help to write such a test.

butterknife-runtime/src/test/java/butterknife/InheritanceTest.java
```diff
@@ -0,0 +1,62 @@
...
+    );
+
+    assertAbout(javaSources()).that(asList(testSource))
+        .withCompilerOptions("-Xlint:-processing")
+        .processedWith(new ButterKnifeProcessor())
+        .compilesWithoutWarnings()
+        .and()
+        .generatesSources(bindingTestSource);
+  }
+}
```
stephanenicolas: I have separted the test to check if we support inheritance of a precompiled class. It covers both incremental compilation and support of inheritance across modules. 

The idea is that we now have a new class (right below in the current PR) that is part of the test code itself. Hence it will be compiled, processed by butterknife too and passed on the classpath of the tests when they are run. It mimics the fact that we are relying on a compiled super class, and not to a super class passed to the processor.

sample/base-library/build.gradle
```diff
@@ -0,0 +1,27 @@
+buildscript {
```
stephanenicolas: I have decomposed the library of the sample into a base library and a library itself so that the sample also covers the case of inheritance across module. Feel free to revert if you don't want to.

sample/library/build.gradle
```diff
@@ -23,7 +23,12 @@ android {
 
 dependencies {
   implementation deps.release.runtime
-  annotationProcessor deps.release.compiler
+  implementation project(':sample:base-library')
+
+  //TODO this change is just to demonstrate that the new compiler works well
+  //but should be reversed when the new version is released
+  //annotationProcessor deps.release.compiler
+  annotationProcessor project(':butterknife-compiler')
```
stephanenicolas: I can revert that as soon as you want, I just use it for debugging purposes. I will probably open a ticket soon on the BK repo: in another project we now use composite builds to manage samples and it works great. It allows to have a clean sample and also use the current modules being developed for testing purposes.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -143,10 +149,18 @@
     debuggable = !"false".equals(env.getOptions().get(OPTION_DEBUGGABLE));
 
     typeUtils = env.getTypeUtils();
+    elementUtils = env.getElementUtils();
     filer = env.getFiler();
     try {
-      trees = Trees.instance(processingEnv);
-    } catch (IllegalArgumentException ignored) {
+      //reflection won't be necessary after https://github.com/gradle/gradle/pull/8393
```
oldergod: The issue is closed, is the comment still relevant?

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
```diff
@@ -371,6 +397,33 @@
     return bindingMap;
   }
 
+  private BindingSet createStubBindingSet(TypeElement parentType) {
+    BindingSet parentBinding;
+    BindingSet.Builder parentBuilder = BindingSet.newBuilder(parentType);
+    if (hasViewBindings(parentType)) {
+      //add a fake field to the parent class so that it will indicate it has a view bindings.
+      //this is required for the subclass to generate a proper view binder
```
oldergod: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

butterknife-runtime/src/test/java/butterknife/BindViewTest.java
```diff
@@ -3,14 +3,22 @@
...
+import javax.lang.model.element.Element;
 import javax.tools.JavaFileObject;
 import javax.tools.StandardLocation;
+
 import org.junit.Test;
 
+import static butterknife.TestStubs.ANDROIDX_CONTEXT_COMPAT;
+import java.util.List;
+import java.util.Map;
+
```
oldergod: not sure imports order is right, and no need to add empty lines between them

sample/library/build.gradle
```diff
@@ -23,7 +23,12 @@ android {
 
 dependencies {
   implementation deps.release.runtime
-  annotationProcessor deps.release.compiler
+  implementation project(':sample:base-library')
+
+  //TODO this change is just to demonstrate that the new compiler works well
+  //but should be reversed when the new version is released
+  //annotationProcessor deps.release.compiler
```
oldergod: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
stephanenicolas: I think it still provides some context. Let me know if you want me to remove it.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
stephanenicolas: I try to do that too. You mean it should start with upper case and end with a `.` ?

butterknife-runtime/src/test/java/butterknife/BindViewTest.java
stephanenicolas: What do you think of applying the spotless plugin in ButterKnife ? I can file a ticket for that.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
oldergod: yes

butterknife-runtime/src/test/java/butterknife/BindViewTest.java
oldergod: Jake can decide, I don't know. My point is rather that whatever rule was applied before is the correct way to organize imports.

butterknife-compiler/src/main/java/butterknife/compiler/ButterKnifeProcessor.java
oldergod: If the issue isn't getting fixed, maybe we can change it. What would future dev would like to see here?

stephanenicolas:
I get your points Benoit. I am open to change my comments.  But I am
waiting to hear from maintainers about the core stuff of this PR.
Thx for the review though.

oldergod:
_Tout est possible à celui qui croit_ and I believe in Jack Wharton.

Alonexx:
This change is not compatible with kapt, since Trees via reflection became invalid again. https://github.com/JetBrains/kotlin/blob/master/plugins/kapt3/kapt3-base/src/org/jetbrains/kotlin/kapt3/base/incremental/incrementalProcessors.kt

dhruvj:
When will this be merged? As Google is recommending everyone to move towards "modular" architecture, I think this problem is only going to get more frequently reported.

stephanenicolas:
@Alone I couldn't find documentation about kapt making it deprecated to
access trees. It is officially already the case in Java with Gradle. But
the changes to make BK incremental do work well, it's a edge case but
everything works fine. I would assume it also works pretty well with kapt.

aritrobanerjee93:
When will this be merged?

Issue #1074, opened 2017-09-11
ObsoleteLintCustomCheck butterknife.lint.InvalidR2UsageDetector 


bakua:
Hi, I have updated our project to Android Gradle plugin 3.0.0-beta5 and now I get the following Lint error:

```xml
<issue 
  id="ObsoleteLintCustomCheck" 
  severity="Error" 
  message="Lint found one or more custom checks using its older Java API; these checks are still run in compatibility mode, but this causes duplicated parsing, and in the next version lint will no longer include this legacy mode. Make sure the following lint detectors are upgraded to the new API: butterknife.lint.InvalidR2UsageDetector" 
  category="Lint" 
  priority="10" 
  summary="Obsolete custom lint check" 
  explanation="Lint can be extended with "custom checks": additional checks implemented by developers and libraries to for example enforce specific API usages required by a library or a company coding style guideline. The Lint APIs are not yet stable, so these checks may either cause a performance, degradation, or stop working, or provide wrong results. This warning flags custom lint checks that are found to be using obsolete APIs and will need to be updated to run in the current lint environment.">
</issue>
```

JakeWharton:
This is already fixed on `master`. You can try 9.0.0-SNAPSHOT if you like.

bakua:
Thank you. Any eta on stable release, please?

JakeWharton:
No ETA

devsparkles:
Cool at least fixed Thanks

devsparkles:
in the next version

PR #1475, opened 2019-02-21
Fix lint check obsolete warning

shaishavgandhi05:


TWiStErRob:
Ref: this is a regression of https://github.com/JakeWharton/butterknife/issues/1074 with a newer version:

> butterknife-runtime-10.1.0.aar\...\lint.jar: Lint found an issue registry (butterknife.lint.LintRegistry) which is older than the current API level; these checks may not work correctly.
> Recompile the checks against the latest version. Custom check API version is 2 (3.2), current lint API level is 3 (3.3+)

@shaishavgandhi05 please add `minApi` to `butterknife/lint/LintRegistry.java`:
```
@Override public int getMinApi() {
    return 2;
}
```
Reason: we know this lint check is compatible with 2 and 3, compiling with 3 is good as it'll make sure this warning goes, but 2 should also be supported, otherwise this PR just moves the warning from one AGP version to another.

PR #1482, opened 2019-03-02
Revert "Integration tests for @OnTouch, @OnItemClick and @OnItemLongClick"

haziminas:
Reverts JakeWharton/butterknife#1477


JakeWharton:
Why?

PR #1483, opened 2019-03-04
Fix BindDimen Float Example on Website

anthonylawson:
Currently, the website shows this as an example of how to bind a `float` dimen:
```java
@BindDimen(R.dimen.spacer) Float spacer; // int (for pixel size) or float (for exact value) field
```
Using `Float` as the type for the `spacer` variable will cause the following error:
```
error: @BindDimen field type must be 'int' or 'float'.
```
The type for the `spacer` variable here should be the primitive `float` type like this:
```java
@BindDimen(R.dimen.spacer) float spacer; // int (for pixel size) or float (for exact value) field
```


JakeWharton:
Thanks. Just deployed.

PR #1486, opened 2019-03-12
Update setup instructions

plastiv:
Include Java8 requirement for the butterknife usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,15 @@ Download
 --------
 
 ```groovy
+android {
+  ...
+  // Butterknife requires Java 8.
+  compileOptions {
+    sourceCompatibility JavaVersion.VERSION_1_8
+    targetCompatibility JavaVersion.VERSION_1_8
+  }
+}
+
 dependencies {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,13 +42,22 @@ Download
...
+}
+
 dependencies {
   implementation 'com.jakewharton:butterknife:10.1.0'
   annotationProcessor 'com.jakewharton:butterknife-compiler:10.1.0'
 }
 ```
 
-If you are using Kotlin, replace `annotationProcessor` with `kapt`.
+Butterknife is AndroidX-enabled and requires `android.useAndroidX=true` in `gradle.properties`. If you are using Kotlin, replace `annotationProcessor` with `kapt`.
```
JakeWharton: This actually isn't true. As of 10.0.0 ButterKnife _only_ supports AndroidX and requires no property to enable this behavior.

plastiv: Thanks for clarification 👍removed ✅ 

I got confused by the wording `and set both of the following Android Gradle plugin flags to true ` here https://developer.android.com/jetpack/androidx#using_androidx .

PR #1520, opened 2019-06-05
Update README.md

DasserBasyouni:
updated ReadMe with a small hint about fixing "Android Gradle plugin 3.0.1 must not be applied to project" error

refrence: https://stackoverflow.com/questions/48245522/android-gradle-plugin-3-0-1-must-not-be-applied-to-project

Have a great day <3

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,10 @@ apply plugin: 'com.android.library'
 apply plugin: 'com.jakewharton.butterknife'
 ```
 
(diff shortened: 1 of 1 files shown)


gh7800:
How do you solve it?

DasserBasyouni:
I don't remember, but I just used the latest butterknife in a library + a project together, do you still have the issue with the latest versions ?

PR #1230, opened 2018-03-24
Support incremental annotation processing.

stephanenicolas:
Addresses issue: #1225 

With #1045, the support for incremental annotation processing is complete.
It would be really nice to see 9.0.0 released with this soon ;)
(let me know if I can help)

PR #1546, opened 2019-08-28
Add support for incremental annotation processing

gavra0:
This commit adds incremental annotation processing support
to Butterknife. More specifically, it makes Butterknife an
isolating annotation processor. More information about what
this means can be found at
https://docs.gradle.org/current/userguide/java_plugin.html#isolating_annotation_processors.

Closes #1230

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ buildscript {
 
       'androidTools': '26.2.0',
       'kotlin': '1.2.71',
+      'incap' : '0.2',
 
       'release': '8.8.1',
   ]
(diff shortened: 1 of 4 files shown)


stephanenicolas:
Does it work incrementally when a subclass is modified but its superclass is unchanged ? 
This was the reason why I had also addressed inerhitance across modules along with incrementality. I might be wrong though..

gavra0:
Support for finding parent bindings in .class files (and classpath) has been added in https://github.com/JakeWharton/butterknife/pull/1545, so that use-case should work.

stephanenicolas:
It's great then. Thx for adding these nice and long awaited features.

PR #1556, opened 2019-10-03
Added Setter sample to implement

harshmittal2810:
Added the setter sample for implementation after `10.2.0` update


JakeWharton:
The readme is not meant to exhaustively cover the API so I would prefer to keep it focused on the binding parts only.

harshmittal2810:
Hi Jake,
But it's not updated on your website too.
That's why I make this pull request.

Issue #1467, opened 2019-02-10
Static imports do not work with butterknife annotation.


mateuszkwiecinski:
I was migrating a project that had been using kapt, butterknife version 8.4.0 and AGP 3.2.1. I wanted to upgrate it to 10.0.0 but it came out that migration to AndroidX is required first, so I ended up with **Butterknife version 9.0.0 and AGP 3.3.1**.
I enountered an runtime error saing butterknife cannot find a view/resource with matching id. (I can find actual error message if needed, but I suppose it was the one that is the most common when using Butterknife).
It came out that in this very particular class arguments passed to `@BindView` annotation **had been statically imported**.

Example:
``` java
import com.app.name.R2.id.some_linear_layout;

@BindView(some_linear_layout) View view;
```
When I've been trying to solve this issue I noticed generated classes for code with static imports didn't have `R.id.` usages and had hardcoded numbers instead:
![image](https://user-images.githubusercontent.com/36954793/52534063-e5e83980-2d3c-11e9-886c-0e824df3aa89.png)

I removed static imports and replaced it with `@BindView(R2.id.some_linear_layout)` (as documentation states) and it started to work again.

I assume this might not be repairable because of changes in annotation processing, but I think it might be worth mentioning that in the documentation.

JakeWharton:
Nah this is fixable, just not accounted for.

dkonurov:
#1562

PR #1562, opened 2019-11-07
fix static import and add test on it

dkonurov:
Closes https://github.com/JakeWharton/butterknife/issues/1467
Try fix this bug. 
Override callback method when visit JCIdent. If JCIdent is Primitive we can detect that it's import static and it's int variable. Add some test's for this


JakeWharton:
Thanks!

PR #1574, opened 2019-12-18
Use fake ID values for R2

### LINKED ISSUE #1549: Every entry of R2 class has ID 0x0 with Android Plugin 3.6.0-alpha10 and Butterknife library plugin
Using 
```
classpath 'com.android.tools.build:gradle:3.6.0-alpha10'
classpath 'com.jakewharton:butterknife-gradle-plugin:10.1.0'
```
and

```
apply plugin: 'com.android.library'
apply plugin: 'com.jakewharton.butterknife'
```

Here is an extrac of the R2 class : 
```
public final class R2 {
  public static final class anim {
    @AnimRes
    public static final int abc_fade_in = 0x0;

    @AnimRes
    public static final int abc_fade_out = 0x0;

    @AnimRes
    public static final int abc_grow_fade_in_from_bottom = 0x0;

    @AnimRes
    public static final int abc_popup_enter = 0x0;

    @AnimRes
    public static final int abc_popup_exit = 0x0;

    @AnimRes
    public static final int abc_shrink_fade_out_from_bottom = 0x0;

    @AnimRes
    public static final int abc_slide_in_bottom = 0x0;

    @AnimRes
    public static final int abc_slide_in_top = 0x0;

    @AnimRes
    public static final int abc_slide_out_bottom = 0x0;

    @AnimRes
    public static final int abc_slide_out_top = 0x0;

    @AnimRes
    public static final int abc_tooltip_enter = 0x0;

    @AnimRes
    public static final int abc_tooltip_exit = 0x0;

    @AnimRes
    public static final int btn_checkbox_to_checked_box_inner_merged_animation = 0x0;

    @AnimRes
    public static final int btn_checkbox_to_checked_box_outer_merged_animation = 0x0;

    @AnimRes
    public static final int btn_checkbox_to_checked_icon_null_animation = 0x0;
```
Reproducible sample : https://github.com/NitroG42/Epoxy6Canary10Issue

If needed I can make an issue on Android Bug Tracker, I'm just not sure if it's a bug or a new behavior that requires some dev in the butterknife plugin.

JakeWharton:
These values are meaningless. We only need them to be unique which is trivially achievable with a counter.

This change is required, however, because AGP 3.6 will stop generating ID values in library projects. All values are set to 0 which obviously will prevent Butter Knife from correctly performing the reverse mapping from R2 value back to name.

Closes #1549.

### DIFF
--- a/butterknife-gradle-plugin/src/test/resources/fixtures/R2.java
+++ b/butterknife-gradle-plugin/src/test/resources/fixtures/R2.java
@@ -20,79 +20,79 @@ import androidx.annotation.StyleableRes;
 public final class R2 {
   public static final class anim {
     @AnimRes
-    public static final int res = 0x7f040001;
+    public static final int res = 1;
   }
 
   public static final class array {
     @ArrayRes
-    public static final int res = 0x7f040002;
+    public static final int res = 2;
   }
 
   public static final class attr {
     @AttrRes
-    public static final int res = 0x7f040003;
+    public static final int res = 3;
   }
 
   public static final class bool {
     @BoolRes
-    public static final int res = 0x7f040004;
+    public static final int res = 4;
   }
 
   public static final class color {
     @ColorRes
-    public static final int res = 0x7f040005;
+    public static final int res = 5;
   }
 
   public static final class dimen {
     @DimenRes
-    public static final int res = 0x7f040006;
+    public static final int res = 6;
   }
 
   public static final class drawable {
     @DrawableRes
-    public static final int res = 0x7f040007;
+    public static final int res = 7;
   }
 
   public static final class id {
     @IdRes
-    public static final int res = 0x7f040008;
+    public static final int res = 8;
   }
 
   public static final class integer {
     @IntegerRes
-    public static final int res = 0x7f040009;
+    public static final int res = 9;
   }
 
   public static final class layout {
     @LayoutRes
-    public static final int res = 0x7f040010;
+    public static final int res = 10;
   }
 
   public static final class menu {
     @MenuRes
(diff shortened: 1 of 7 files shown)


butterknife-gradle-plugin/src/test/resources/fixtures/R2.java
```diff
@@ -20,79 +20,79 @@
...
   public static final class style {
     @StyleRes
-    public static final int res = 0x7f040014;
+    public static final int res = 14;
   }
 
   public static final class styleable {
     @StyleableRes
-    public static final int resArray_child = 0;
+    public static final int resArray_child = 15;
```
imorlowska: BTW, for children of declare styleables it's actually pretty interesting since, if we had:
<resources> 
    <attr name="foo"/>
    <declare-styleable name="bar">
        <item type="attr" name="foo"/> //re-using 'foo' defined above
    </declare-styleable>
</resoures>

In the (real) R class R.attr.foo and R.styleable.bar[R.styleable.bar_foo] should have the same value:
R.attr.foo = 12345
R.styleable.bar = [12345]
R.styleable.bar_foo = 0 // index in the R.styleable.bar array

I don't think this will affect this change at all, but just an FYI. :)

imorlowska: Sorry, the comment got rid of all the sharp brackets... The code that was removed:
```
<resources>
    <attr name="foo"/>
    <declare-styleable name="bar">
        <item type="attr" name="foo"/> //re-using 'foo' defined above
    </declare-styleable>
</resources>
```

Hamdy1:
Beautiful

PR #1620, opened 2020-07-22
Avoid all @OnClick events may not working

### LINKED ISSUE #1619: All methods annotated with @OnClick may not work
Today i found all click methods annotated with `@OnClick` are not working in my code.

After debugging, i found something wrong with `DebouncingOnClickListener`, 
the static field `enabled` always stays in `false`.

```
public abstract class DebouncingOnClickListener implements View.OnClickListener {
  static boolean enabled = true;

  private static final Runnable ENABLE_AGAIN = () -> enabled = true;

  @Override public final void onClick(View v) {
    if (enabled) {
      enabled = false;
      //we can NOT guarantee that ENABLE_AGAIN will be executed.
      v.post(ENABLE_AGAIN);
      doClick(v);
    }
  }

  public abstract void doClick(View v);
}
```

Certain cases can lead to this, here is my case:
 i called `someButton.performClick()` programmatically while `someButton` is not attached to window and will not be attached to window. `DebouncingOnClickListener` of `someButton`  set `enabled` to `false`. 
At this time `ENABLE_AGAIN` will never get chance to be executed.

So, do we have some other `DebouncingOnClickListener` implementations to avoid this?

Thanks in advance.

lonnyzhang423:
Avoid `DebouncingOnClickListener` static field `enabled` staying in `false` state. 

Fixes #1619

### DIFF
--- a/butterknife-runtime/src/main/java/butterknife/internal/DebouncingOnClickListener.java
+++ b/butterknife-runtime/src/main/java/butterknife/internal/DebouncingOnClickListener.java
@@ -1,5 +1,7 @@
 package butterknife.internal;
 
+import android.os.Handler;
+import android.os.Looper;
 import android.view.View;
 
 /**
@@ -7,14 +9,20 @@ import android.view.View;
  * same frame. A click on one button disables all buttons for that frame.
  */
 public abstract class DebouncingOnClickListener implements View.OnClickListener {
-  static boolean enabled = true;
-
   private static final Runnable ENABLE_AGAIN = () -> enabled = true;
+  private static final Handler MAIN = new Handler(Looper.getMainLooper());
+
+  static boolean enabled = true;
 
   @Override public final void onClick(View v) {
     if (enabled) {
       enabled = false;
-      v.post(ENABLE_AGAIN);
+
+      // Post to the main looper directly rather than going through the view.
+      // Ensure that ENABLE_AGAIN will be executed, avoid static field {@link #enabled}
+      // staying in false state.
+      MAIN.post(ENABLE_AGAIN);
+
       doClick(v);
     }
   }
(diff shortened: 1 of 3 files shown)


lonnyzhang423:
Fixes issue #1619

butterknife-runtime/src/main/java/butterknife/internal/DebouncingOnClickListener.java
```diff
@@ -13,8 +15,14 @@
 
   @Override public final void onClick(View v) {
     if (enabled) {
-      enabled = false;
-      v.post(ENABLE_AGAIN);
```
JakeWharton, line 17: We can switch this to post to the main looper directly rather than going through the view. That way we don't need to check for attach.

lonnyzhang423: Thanks for your reply.

Yes, i updated my code and pushed to my fork. But the Travis CI build failed, i don't known why...

CI failed log:

```
Task :butterknife-integration-test:connectedCodegenDebugAndroidTest FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':butterknife-integration-test:connectedCodegenDebugAndroidTest'.
> There were failing tests. See the report at: file:///home/travis/build/JakeWharton/butterknife/butterknife-integration-test/build/reports/androidTests/connected/flavors/CODEGEN/index.html
```

lonnyzhang423: Hi Jake, I fixed unit test errors and The Travis CI build passed😀.

JakeWharton:
I changed the test so that it's black box instead of mucking with the internals through reflection. Thanks! Will do a release now.

lonnyzhang423:
Well done, thanks.

PR #1622, opened 2020-08-05
Fix IllegalArgumentException in @Touch in butterknife-reflect

lernerb:
Looks like we aren't passing through the MotionEvent.class when using butterknife-reflect.

> java.lang.IllegalArgumentException: Wrong number of arguments; expected 2, got 1

While this shouldn't effect any production builds since folks shouldn't be using this in prod, it can cause `@OnTouch` annotations to crash when using in debug builds for users that use it.


JakeWharton:
Would be nice to have a test, but whevs. This project isn't getting large-scale refactorings such that this would ever break.

JakeWharton:
Test in https://github.com/JakeWharton/butterknife/pull/1623

PR #1627, opened 2020-08-31
Maybe we can

ThirdGoddess:
Because ’parameterTypes.length‘ It is a value in itself. If you write as before, it may reduce the clarity of the code！So I think it can be modified in this way！


JakeWharton:
This library is in maintenance mode. Only bugfixes will be merged.

In general, this changes the behavior though to query the length on every iteration. You can confirm this by looking at the change in Java bytecode. It's possible (or likely) that R8 would hoist the length lookup outside the loop, but I would have to confirm that and not everyone runs R8 on their apps.

